PR #20, opened 2021-03-31
Simple hello world using refl-cpp

patricia-gallardo:
This adds a concepts folder for refl-cpp and the cmake integration for fetching refl-cpp.
The folder contains a small example, and the CMakeLists.txt shows how to use the refl-cpp made available through CMake.

### DIFF
--- a/cmake/CompilerWarnings.cmake
+++ b/cmake/CompilerWarnings.cmake
@@ -47,6 +47,7 @@ function(set_project_warnings project_name)
       -Wnull-dereference # warn if a null dereference is detected
       -Wdouble-promotion # warn if float is implicit promoted to double
       -Wformat=2 # warn on security issues around functions that format output (ie printf)
+      -Wno-gnu-zero-variadic-macro-arguments # allow zero variadic macro arguments (for refl_cpp)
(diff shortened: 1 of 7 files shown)


cmake/CompilerWarnings.cmake
```diff
@@ -47,6 +47,7 @@ function(set_project_warnings project_name)
       -Wnull-dereference # warn if a null dereference is detected
       -Wdouble-promotion # warn if float is implicit promoted to double
       -Wformat=2 # warn on security issues around functions that format output (ie printf)
+      -Wno-gnu-zero-variadic-macro-arguments # allow zero variadic macro arguments (for refl_cpp)
```
wirew0rm, line 50: For me it also builds without this line, possibly also because refl-cpp is now a system include?

wirew0rm (approved):
Very nice, looking forward to how this will evolve.

cmake/CompilerWarnings.cmake
patricia-gallardo: It's a mac only warning suppression, so it builds on windows and linux without it

PR #22, opened 2021-04-06
Hello world in rttr

patricia-gallardo:


patricia-gallardo:
Moving this work over to the concepts repo

PR #34, opened 2021-06-01
Coverage

wirew0rm:
Add coverage reporting to the CI.

A local coverage html coverage report can be generated with the coverage_html cmake target, while the coverage target will generate a cobertura xml report which is used by the CI to upload to codecov and codacy.


RalphSteinhagen (approved):
lgtm and as discussed. :+1:

PR #38, opened 2021-07-28
simplify cmake structure and add conan buildfile

wirew0rm:
This allows to consume the library from any conan enabled project and install it system wide. For the non conan case, the target names and package discovery will have to be changed to allow for that, too. (probably add find_package calls to opencmwConfig.cmake.in template and create CONAN_PKG:: target aliases for their targets).

- Use INTERFACE targets instead of globally adding include directories.
- move functionalities into different subdirectories in src/
- add install targets
- add enable flag for tests and disable them in conan build
- make opencmw a conan package itself:
  `mkdir conanbuild && cd conanbuild && conan install ..`
  -> add requires = opencmw/0.0.1 to your package
  not in conan-center yet, first do some iterations locally


RalphSteinhagen (commented):
LGTM!

PR #57, opened 2021-11-19
made MimeType and helper functions constexpr capable

RalphSteinhagen:
updated MimeType to support `constexpr` evaluation ... also needed `static_for(..)` definition ... 

It's still WIP w.r.t. API naming of functions but open for comments...

### DIFF
--- a/src/core/include/MIME.hpp
+++ b/src/core/include/MIME.hpp
@@ -5,28 +5,29 @@
 #include <array>
 #include <cctype> // for tolower
 #include <iostream>
+#include <span>
 #include <string_view>
 #include <vector>
 
 namespace opencmw::MIME {
 
 class MimeType {
-    static constexpr std::string_view   space = " ";
-    const std::string_view              _typeName;
-    const std::string_view              _description;
-    const std::vector<std::string_view> _fileExtensions;
+    const std::string_view                _typeName;
+    const std::string_view                _description;
+    const std::array<std::string_view, 3> _fileExtensions;
+    const size_t                          _N;
 
 public:
     MimeType() = delete;
     template<class... Ts>
     constexpr explicit MimeType(const char *name, const char *description, Ts... ext) noexcept
-        : _typeName(name), _description(description), _fileExtensions{ std::forward<Ts>(ext)... } {}
-    const std::string_view              typeName() const noexcept { return _typeName; }
-    const std::string_view              description() const noexcept { return _description; }
-    const std::vector<std::string_view> fileExtensions() const noexcept { return _fileExtensions; }
+        : _typeName(name), _description(description), _fileExtensions{ std::forward<Ts>(ext)... }, _N(sizeof...(Ts)) {}
+    constexpr std::string_view                  typeName() const noexcept { return _typeName; }
+    constexpr std::string_view                  description() const noexcept { return _description; }
+    constexpr std::span<const std::string_view> fileExtensions() const noexcept { return std::span(_fileExtensions.data(), _N); };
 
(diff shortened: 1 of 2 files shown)


src/core/include/MIME.hpp
```diff
@@ -117,50 +118,92 @@ static inline std::array ALL{
...
+                index = i;
+            }
+        });
+        return (index == SIZE_MAX) ? UNKNOWN : ALL[index];
+    }
+
     for (const MimeType &type : ALL) {
         // N.B.mimeType may contain several MIME types, e.g "image/webp,image/apng,image/*"
-        constexpr auto lowerCaseCompare = [](char ch1, char ch2) noexcept { return std::tolower(ch1) == std::tolower(ch2); };
+        constexpr auto lowerCaseCompare = [](const char ch1, const char ch2) noexcept -> bool { return detail::toLower(ch1) == detail::toLower(ch2); };
```
ivan-cukic, line 159: This is a useful function (similar lambda exists in `majordomo/Utils.hpp` - might be worth it to move it to a separate header with all case-insensitive string stuff that we end up needing.

src/core/include/MIME.hpp
```diff
@@ -5,28 +5,29 @@
...
 namespace opencmw::MIME {
 
 class MimeType {
-    static constexpr std::string_view   space = " ";
-    const std::string_view              _typeName;
-    const std::string_view              _description;
-    const std::vector<std::string_view> _fileExtensions;
+    const std::string_view                _typeName;
+    const std::string_view                _description;
+    const std::array<std::string_view, 3> _fileExtensions;
```
ivan-cukic, line 17: If this is going to end up as something else than an array, please add a `TODO:`. If not, it would be nice if we had a named constant that equals 3 (something like `MAX_MIMETYPE_EXTENSIONS`)

ivan-cukic (commented):
No big issues.

Shame something like [1] is not in `std::`, would make some things easier.

[1] https://github.com/bolero-MURAKAMI/Sprout/

src/core/include/MIME.hpp
RalphSteinhagen: yeah, was wondering about that ... hoped to find a `std::` `constexpr` function for compare-ignoring-case.

I'd wait a bit before making this a proper API since I learned that C++20's `constexpr std::string` is just around the corner for gcc and hopefully also clang. Once these are officially released, I'd also refactor the string_views to std::strings again for which there are more of these conversion/comparison functions....

src/core/include/MIME.hpp
RalphSteinhagen: These are basically only compile-time `constexpr` definitions that probably never change ... can add a comment to this magical number on why only 3

RalphSteinhagen: added TODO comment:
``` 
// N.B. max number '3' of ext is hardcoded TODO: replace with constexpr std::array or std::vector once fully supported by both gcc13 and clang?
```

src/core/include/MIME.hpp
RalphSteinhagen: added TODO comment:
``` 
// TODO: replace with case-insensitive constexpr std::string once fully supported by both gcc13 and clang?, see http://www.gotw.ca/gotw/029.htm
```
This is a nominal C++20 functionality: gcc has it already implemented in trunk (gcc13) already, clang is pending
N.B. This refactoring should also include the various uses of `refl::util::const_string' in both refl-cpp as well as mp-units

wirew0rm (approved):
:+1:

PR #64, opened 2021-12-03
Fix undefined behavior on URI copy

ivan-cukic:


RalphSteinhagen (approved):
very good and concise fix (copy constructor didn't adjust the string_view references). :+1:

PR #78, opened 2021-12-08
Enable coverage explicitly

wirew0rm:
Make ENABLE_COVERAGE default to OFF and actively enable it in the github action's debug/coverage build.
Enabling coverage for non debug builds does not lead to useful results and for the debugging usecase coverage is also not required.

Thanks to @frankosterfeld for the suggestion.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,7 +63,7 @@ include(cmake/ReflCpp.cmake) # todo: switch to obtaining reflcpp from conan
 if(ENABLE_TESTING)
   enable_testing()
   message("Building Tests.")
-  if (ENABLE_COVERAGE)
+  if (ENABLE_COVERAGE AND (CMAKE_BUILD_TYPE STREQUAL "Debug" OR GENERATOR_IS_MULTI_CONFIG))
     if(UNIX AND NOT APPLE) # Linux
       message("Coverage reporting enabled")
       include(cmake/CodeCoverage.cmake) # https://github.com/bilke/cmake-modules/blob/master/CodeCoverage.cmake (License: BSL-1.0)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -63,7 +63,7 @@ include(cmake/ReflCpp.cmake) # todo: switch to obtaining reflcpp from conan
 if(ENABLE_TESTING)
   enable_testing()
   message("Building Tests.")
-  if (ENABLE_COVERAGE)
+  if (ENABLE_COVERAGE AND (CMAKE_BUILD_TYPE STREQUAL "Debug" OR GENERATOR_IS_MULTI_CONFIG))
```
frankosterfeld, line 66: Thinking about this, I find it slightly confusing: I could pass `-DENABLE_COVERAGE=ON` and still get no coverage analysis, if the buildtype isn't Debug nor multi-config. (and with multi-config, I would get a release build without optimizations?). I'd tend to let ENABLE_COVERAGE default to OFF: if one doesn't run coverage tests, well, and if one does, one will notice soon enough and add the flag.

wirew0rm: That was my first impulse, too. But after discussion with @RalphSteinhagen and trying it out in clion I came to the conclusion, that default OFF might look cleaner, but gets more in the way in daily use. E.g. if you hit coverage in clion, it will promt you to create a new CMakeProfile enabling a different set of coverage options, whereas with default ON it just works. Release builds with coverage created a big warning before not to use them. Also the documentation of the ENABLE_COVERAGE flag states that it s only effective on Debug builds (it says test builds but that should be fixed too). The multi generator bit is copied from the coverage cmake file, so I guess that they handle this usecase separately.

wirew0rm: Changed it so just the default gets changed depending on the build type.

frankosterfeld (approved):

PR #82, opened 2021-12-13
Fix build with CMake 3.18

frankosterfeld:
Specifying source files is a CMake 3.19 feature.


frankosterfeld:
Not sure what the exact intention here was, so please check if this is acceptable :)

wirew0rm (approved):
The current state does not install the fast_float headers when using the cmake install target. I'm not completely sure if this line has any effect, as the line below is adding the include directory as a whole anyway.
I also found some old filename references when trying the install step, which i fixed in a separate commit in this branch.
I would suggest merging this and later investigating if we want to make use of this new 3.19 feature (and bump the minimal version accordingly).

PR #83, opened 2021-12-13
Set all tests as dependencies for coverage targets

frankosterfeld:
Make sure majordomo_tests and core_tests are (re-)built before
coverage/coverage_html is run.


wirew0rm (approved):
Looks good :+1:

PR #86, opened 2021-12-14
Frank/query pubsub

frankosterfeld:
Consider query parameters when matching subscriptions.

WIP because it contains some hack^Whotfixes for URI that Ralph might want to fix differently/properly.


ivan-cukic (approved):

RalphSteinhagen (approved):
LGTM! :+1: 
API changes were incorporated as discussed.

PR #90, opened 2021-12-15
Update codecov-action to v2

wirew0rm:
v1 uses the deprecated bash uploader and is not maintained anymore. This also might fix some spurious failed coverage uploads we have [observed](https://github.com/fair-acc/opencmw-cpp/pull/83#issuecomment-994824884).

All other github actions are up to date.


RalphSteinhagen (approved):
LGTM :+1:

PR #92, opened 2021-12-17
thread affinity and priority utils

RalphSteinhagen:
N.B. assumes UNIX/POSIX threads for the time being, other OSes are nullopts

* non-root users may need SYS_CAP_NICE capabilities being set for the given application in order to make use of the FIFO and ROUND_ROBIN scheduling policies:

sudo setcap 'cap_sys_nice=eip' <application>

### DIFF
--- a/src/core/include/ThreadAffinity.hpp
+++ b/src/core/include/ThreadAffinity.hpp
@@ -162,8 +162,8 @@ constexpr cpu_set_t getAffinityMask(const T &threadMap) {
 #endif
 } // namespace detail
 
-std::vector<bool> getThreadAffinity(thread_type auto &...thread) {
 #ifdef _POSIX_VERSION
+std::vector<bool> getThreadAffinity(thread_type auto &...thread) {
     const pthread_t handle = detail::getPosixHandler(thread...);
     if (handle == 0U) {
         throw std::system_error(THREAD_UNINITIALISED, thread_exception(), fmt::format("getThreadAffinity(thread_type)"));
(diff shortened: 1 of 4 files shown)


RalphSteinhagen:
N.B. introducing a 'threadGuard' within Catch2 that is not thread-safe doesn't seem to be a very smart solution. The built-in Catch2 test time-out seems to work sufficiently well.

wirew0rm (approved):
Looks good, I think the test coverage is also appropriate given that most of the misses are error conditions that cannot be easily triggered on the CI.

src/core/include/ThreadAffinity.hpp
```diff
@@ -0,0 +1,330 @@
...
+
+    const char *name() const noexcept { return "thread_exception"; };
+    std::string message(int errorCode) const {
+        switch (errorCode) {
+        case THREAD_UNINITIALISED:
+            return "thread uninitialised or user does not have the appropriate rights (ie. CAP_SYS_NICE capability)";
+        case THREAD_ERROR_UNKNOWN:
+            return "thread error code 2";
+        case THREAD_ERANGE:
+            return fmt::format("length of the string specified pointed to by name exceeds the allowed limit THREAD_MAX_NAME_LENGTH = '{}'", THREAD_MAX_NAME_LENGTH);
```
frankosterfeld, line 45: This one should be easy to add coverage for?

src/core/include/ThreadAffinity.hpp
```diff
@@ -0,0 +1,330 @@
...
+            CPU_CLR(i, &cpuSet);
+        }
+    }
+    return cpuSet;
+}
+#endif
+} // namespace detail
+
+std::vector<bool> getThreadAffinity(thread_type auto &...thread) {
+#ifdef _POSIX_VERSION
```
frankosterfeld, line 166: won't compilation fail with -Werror because of unused parameters here, on non-posix?

src/core/include/ThreadAffinity.hpp
RalphSteinhagen: added additional mini-unit tests for these

wirew0rm (approved):
:+1:

PR #95, opened 2021-12-17
YaS Serialiser addons

RalphSteinhagen:
* fixed IoBuffer with JSON content tripping the YaS protocol
* added additional checks against potential out-of-bound reads for bulk operations in IoBuffer
* changed throwing exception to skipping writing into `const` marker fields

work-in-progress:
 * adding (de-)Serialiser definitions for std::unordered_map and other primitives already implemented on the Java side

### DIFF
--- a/src/serialiser/include/IoSerialiserYaS.hpp
+++ b/src/serialiser/include/IoSerialiserYaS.hpp
@@ -57,7 +57,6 @@ template<> inline constexpr uint8_t getDataTypeId<std::string_view[]>() { return
 // template<> inline constexpr uint8_t getDataTypeId<START_MARKER>() { return 200; }
 // template<> inline constexpr uint8_t getDataTypeId<enum>()          { return 201; }
 // template<typename T> inline constexpr uint8_t getDataTypeId<std::list<T>>()     { return 202; }
-// template<> inline constexpr uint8_t getDataTypeId<std::unsorted_map>() { return 203; }
 // template<> inline constexpr uint8_t getDataTypeId<std::queue>()    { return 204; }
 // template<> inline constexpr uint8_t getDataTypeId<std::set>()      { return 205; }
 
@@ -78,7 +77,7 @@ struct IoSerialiser<YaS, T> {
 template<Number T> // catches all numbers
 struct IoSerialiser<YaS, T> {
     inline static constexpr uint8_t getDataTypeId() { return yas::getDataTypeId<T>(); }
-    constexpr static bool           serialise(IoBuffer &buffer, const ClassField & /*field*/, const T &value) noexcept {
+    constexpr static bool           serialise(IoBuffer &buffer, const ClassField           &/*field*/, const T &value) noexcept {
         buffer.put(value);
         return std::is_constant_evaluated();
     }
@@ -91,7 +90,7 @@ struct IoSerialiser<YaS, T> {
 template<StringLike T>
 struct IoSerialiser<YaS, T> {
     inline static constexpr uint8_t getDataTypeId() { return yas::getDataTypeId<T>(); }
-    constexpr static bool           serialise(IoBuffer &buffer, const ClassField & /*field*/, const T &value) noexcept {
+    constexpr static bool           serialise(IoBuffer &buffer, const ClassField           &/*field*/, const T &value) noexcept {
         buffer.put<T>(value); // N.B. ensure that the wrapped value and not the annotation itself is serialised
         return std::is_constant_evaluated();
     }
@@ -104,7 +103,7 @@ struct IoSerialiser<YaS, T> {
 template<ArrayOrVector T>
 struct IoSerialiser<YaS, T> {
     inline static constexpr uint8_t getDataTypeId() { return yas::getDataTypeId<T>(); }
-    constexpr static bool           serialise(IoBuffer &buffer, const ClassField & /*field*/, const T &value) noexcept {
+    constexpr static bool           serialise(IoBuffer &buffer, const ClassField           &/*field*/, const T &value) noexcept {
         buffer.put(std::array<int32_t, 1>{ static_cast<int32_t>(value.size()) });
         buffer.put(value);
         return std::is_constant_evaluated();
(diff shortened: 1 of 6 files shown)


src/serialiser/include/IoSerialiserYaS.hpp
```diff
@@ -148,11 +147,110 @@ struct IoSerialiser<YaS, T> {
     }
 };
 
+template<MapLike T>
+struct IoSerialiser<YaS, T> {
+    inline static constexpr uint8_t getDataTypeId() { return 203; }
```
wirew0rm, line 152: Magic constant instead of `yas::getDataTypeId<...>()` because of partial specialisation problem? Does it make sense to remove the yas::getDataTypeId block completely if it cannot be used everywhere? (cmwlight serialiser also has this).
Alternatively comment here why this is used as a Literal value and add a comment in the `yas::` block above s.t. when reading it is clear why it is missing?

src/serialiser/include/IoSerialiserYaS.hpp
```diff
@@ -148,11 +147,110 @@ struct IoSerialiser<YaS, T> {
...
+            buffer.put(yas::getDataTypeId<OTHER>()); // type-id
+            buffer.put(typeName<K>());               // primary type
+            buffer.put("");                          // secondary type (if any) TODO: add appropriate
+            buffer.put(nElements);
+            for (auto &entry : value) {
+                IoSerialiser<YaS, K>::serialise(buffer, field, entry.first);
+            }
+        }
+
+        if constexpr (is_supported_number<V> || is_stringlike<V>) {
```
wirew0rm, line 178: what about double[]? it has it's own type id, so it should probably by serialised here?

src/serialiser/include/IoSerialiserYaS.hpp
```diff
@@ -148,11 +147,110 @@ struct IoSerialiser<YaS, T> {
...
+        using V                 = typename T::mapped_type;
+        const int32_t nElements = static_cast<int32_t>(value.size());
+        buffer.put(std::array<int32_t, 1>{ nElements }); // [ndims]{size}
+        buffer.put(nElements);                           // nElements
+
+        if constexpr (is_supported_number<K> || is_stringlike<K>) {
+            constexpr int entrySize = 17; // as an initial estimate
+            buffer.reserve_spare(static_cast<size_t>(nElements * entrySize + 9));
+            buffer.put(static_cast<uint8_t>(yas::getDataTypeId<K>())); // type-id
+            buffer.put(nElements);                                     // nElements
```
wirew0rm, line 164: I would opt for not storing the number of elements redundantly, just takes up space and needs code to validate it again that should never be triggered when serialising correctly. Maybe also move the Types K and V before the data, that way, we can fail early in the deserialiser and do not have to deserialise all the keys just to flunk out on a value type mismatch.

src/serialiser/include/IoSerialiserYaS.hpp
```diff
@@ -148,11 +147,110 @@ struct IoSerialiser<YaS, T> {
...
+            for (auto &entry : value) {
+                buffer.put(entry.second);
+            }
+        } else {                                         // non-primitive or non-string-like types
+            buffer.put(yas::getDataTypeId<OTHER>());     // type-id
+            buffer.put(typeName<V>());                   // primary type
+            buffer.put("");                              // secondary type (if any) TODO: add appropriate
+            buffer.put(static_cast<int32_t>(nElements)); // nElements
+            for (auto &entry : value) {
+                IoSerialiser<YaS, K>::serialise(buffer, field, entry.second);
```
wirew0rm, line 192: I think there is no IoSerialiser<YaS, ReflectableClass> serialiser implemented, this would have to call serialise<YaS, entry.second, hierarchyLevel) to use the reflection based serialiser here. Or we would have to unify the interface and implement the mentioned IoSerialiser for ReflectableClass, which I think would be nice but would be a big refactor (including pulling metaInformation handlers and hierarchy depth into the primitive Serialisers, I don't know if that would influence performance). Also this would write the Type name into every Element's header instead of once per set (could of course be changed/refactored).
But since the deserialiser is also not implemented, this could be deferred for later. If this code is currently not used, I would suggest to comment it out or remove it.

src/serialiser/include/IoSerialiserYaS.hpp
```diff
@@ -57,7 +57,6 @@ template<> inline constexpr uint8_t getDataTypeId<std::string_view[]>() { return
 // template<> inline constexpr uint8_t getDataTypeId<START_MARKER>() { return 200; }
 // template<> inline constexpr uint8_t getDataTypeId<enum>()          { return 201; }
 // template<typename T> inline constexpr uint8_t getDataTypeId<std::list<T>>()     { return 202; }
-// template<> inline constexpr uint8_t getDataTypeId<std::unsorted_map>() { return 203; }
```
wirew0rm, line 60: see below: maybe leave a comment that this is implemented and why it is not added here with the other definitions.

src/serialiser/include/IoSerialiserYaS.hpp
```diff
@@ -148,11 +147,110 @@ struct IoSerialiser<YaS, T> {
...
+
+        if constexpr (is_supported_number<V> || is_stringlike<V>) {
+            constexpr int entrySize = 17; // as an initial estimate
+            buffer.reserve_spare(static_cast<size_t>(nElements * entrySize + 9));
+            buffer.put(static_cast<uint8_t>(yas::getDataTypeId<V>()));
+            buffer.put(nElements);
+            for (auto &entry : value) {
+                buffer.put(entry.second);
+            }
+        } else {                                         // non-primitive or non-string-like types
```
wirew0rm, line 186: what exactly is supposed to be in this branch? `Map<T, Map/Vector/ReflectableClass/...>` TypeId OTHER hints to Reflectable class, as List/Map/etc have their own type IDs.

wirew0rm (approved):
The fixes look all good, for the map serialisation I made a few suggestions, let me know what you think and feel free to incorporate them or disagree with them. I think map is a really important addition to YaS!

src/serialiser/include/IoSerialiserYaS.hpp
RalphSteinhagen: I used the literal value because of `std::[unordered_]map` didn't allow for partial specialisation. Should work -- in theory -- be the same as `std::string` but for some reason it didn't work. I see this as an exception to the rule but we equally could inline these 'magic numbers' for the other use-cases since the locality of the definition is anyway given...

src/serialiser/include/IoSerialiserYaS.hpp
RalphSteinhagen: Do we want to support c-style arrays? The handling is a bit different than in Java since raw-arrays are length-limited and behave a bit more like vector<double>. The implemented function is -- to first order -- functionally identical to the one in opencwm-java. If we want to extend this we should do this on both sides.

src/serialiser/include/IoSerialiserYaS.hpp
RalphSteinhagen: I agree, but I copy-pasted this to be compatible with the existing opencmw-java implentation.

src/serialiser/include/IoSerialiserYaS.hpp
RalphSteinhagen: This would be a new functionality that goes beyond this little basic PR.  My aim was not to have a feature complete map support but rather to cover the most common use-cases that we already use (e.g. the DataSet interace storing int, strings, etc. within the style maps, etc.)

Feel free to add extensions. :smiley:

src/serialiser/include/IoSerialiserYaS.hpp
RalphSteinhagen: this is a carbon copy of the opencmw-java implementation. The OTHER hints at that these are user-supplied serialisers/deserialisers for which there are no specific type ID defined.

src/serialiser/include/IoSerialiserYaS.hpp
RalphSteinhagen: also `double[]` is not really a primitive type and nesting of types with arbitrary recursion depth would need quite a bit more polished handling of deallocating/allocating objects etc.

PR #98, opened 2021-12-25
simplified and clean-up of IoSerialiser::[de]serialiser functions

RalphSteinhagen:
* simplified concept declarations in opencmw.hpp and added new macros/helper functions    
  - `#define forceinline inline __attribute__((always_inline))`
    -> use this for hot-spots only <-> may bloat code size, not fit into cache and consequently slow down execution
    N.B. compatible with gcc, clang and icc nominal C++ 'inline' keyword is only a suggestion for these compilers
    and Boosted the YaS serialiser performance from  ~ 7.7 -> 8.5 GB/s
  - `#define neverinline __attribute__((noinline))`
    -> primarily used to avoid inlining (rare) exception handling code
  - `#define FWD(...) ::std::forward<decltype(__VA_ARGS__)>(__VA_ARGS__)`
    -> short-hand notation for easier/frequent perfect forwarding use cases (e.g. `auto &&`)
  - `auto &unmove(auto &&t) { return t; }` -> opposite of std::move(...)

* merged IoBuffer 'putRaw' into templatized 'put' w/ same functionality
  N.B. rationale: reduced the number of member function signatures

* (tried to) reduce cognitive complexity by:
  - introducing (hopefully) more meaningful function names
  - refactored code into new `handleDeserialisationErrorAndSkipToNextField(..)`
  - removed redundant code documentation for function headers (didn't add to enhance the understanding what the function does beyond naming the function parameters)
  - removed superfluous ::serialiser bool return-type (N.B. was only used for debugging constexpr-ness)
  - replaced multi-parameter function arguments with FieldDescription[Short, Long]
    N.B. this is also needed in view of a YAML/OpenAPI where the individual field/type serialiser need the hierarchy depth and (if requested) other meta information


wirew0rm (approved):
:+1:

PR #102, opened 2022-01-04
WIP terminate on invalid JSON

frankosterfeld:
Before, deserialise would not terminate on "{ "id" : 42 ]".


frankosterfeld:
Not sure this is correct (it was a first shot in the dark), but the json serialiser tests pass this way...

PR #105, opened 2022-01-05
Add RBAC priority handling; and mmi.dns, mmi.openapi and mmi.echo

frankosterfeld:


RalphSteinhagen (approved):
changes made as discussed via matrix.io. LGTM! :+1:

@frankosterfeld thanks for your infinite patience with this long PR and API iterations.
The Broker/Worker API is in nice shape now!

PR #111, opened 2022-01-17
CmwLight serialiser

wirew0rm:
(De-)serialisation for the CMW middleware wire protocol.

TODO:
- [x] treatment of missing fields/domain object differences (different from other protocols because there is no easy way to skip fields, one has to keep deserialising, ignore the result and count the number of deserialised fields)

### DIFF
--- a/src/serialiser/include/opencmw.hpp
+++ b/src/serialiser/include/opencmw.hpp
@@ -30,6 +30,9 @@ inline constexpr const bool is_quantity<T> = true;
 } // namespace units::detail
 
 namespace opencmw {
+template<typename>
+constexpr bool always_false = false;
+
 using units::basic_fixed_string;
 using units::is_same_v;
 
--- a/src/serialiser/include/opencmw.hpp
+++ b/src/serialiser/include/opencmw.hpp
@@ -48,6 +48,8 @@ inline constexpr bool isReflectableClass() {
     }
     return false;
 }
+template<typename T>
+constexpr bool isReflectableClass_v = isReflectableClass<T>();
 template<class T>
 concept ReflectableClass = isReflectableClass<T>();
 
--- a/.gitignore
+++ b/.gitignore
@@ -24,8 +24,7 @@ imgui.ini
 bin/
 build/
 out/
-cmake-build-debug/
-cmake-build-release/
+cmake-build-*/
 
 # Annoying OS Generated Files
 .DS_Store
--- a/concepts/serialiser/IoSerialiserBenchmark.cpp
+++ b/concepts/serialiser/IoSerialiserBenchmark.cpp
@@ -1,4 +1,5 @@
 #include "IoSerialiserBenchmark.hpp"
+#include <IoSerialiserCmwLight.hpp>
 #include <IoSerialiserJson.hpp>
 #include <IoSerialiserYAML.hpp>
 #include <IoSerialiserYaS.hpp>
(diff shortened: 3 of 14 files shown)


RalphSteinhagen (approved):
LGTM! :+1:

RalphSteinhagen:
for reference: before/after serialiser performance comparison
```
<main> (2 x 10 runs)
┌─protocol─┬────────ALWAYS─────────┬────────LENIENT────────┬────────IGNORE─────────┐
│   YAML   │ 99.4 MB/s ±   3.2 MB/s│ 99.8 MB/s ±   2.9 MB/s│102.0 MB/s ±   1.1 MB/s│
│   Json   │330.5 MB/s ±   3.7 MB/s│328.3 MB/s ±   4.4 MB/s│322.9 MB/s ± 638.5 kB/s│
│   YaS    │  6.6 GB/s ±  23.4 MB/s│  6.5 GB/s ±   4.9 MB/s│  7.5 GB/s ±   9.6 MB/s│
└──────────┴───────────────────────┴───────────────────────┴───────────────────────┘

┌─protocol─┬────────ALWAYS─────────┬────────LENIENT────────┬────────IGNORE─────────┐
│   YAML   │100.7 MB/s ± 584.8 kB/s│100.2 MB/s ±  90.8 kB/s│100.7 MB/s ±  70.3 kB/s│
│   Json   │326.1 MB/s ± 526.5 kB/s│325.6 MB/s ± 639.0 kB/s│326.6 MB/s ± 483.9 kB/s│
│   YaS    │  6.8 GB/s ±  23.5 MB/s│  6.6 GB/s ±   8.8 MB/s│  7.6 GB/s ±   7.2 MB/s│
└──────────┴───────────────────────┴───────────────────────┴───────────────────────┘

<cmwLightSerialiser> (3 x 10 runs)
┌─protocol─┬────────ALWAYS─────────┬────────LENIENT────────┬────────IGNORE─────────┐
│   YAML   │102.3 MB/s ±   2.3 MB/s│100.9 MB/s ± 811.8 kB/s│101.7 MB/s ± 195.2 kB/s│
│   Json   │326.1 MB/s ± 525.6 kB/s│324.5 MB/s ± 594.2 kB/s│325.7 MB/s ± 570.8 kB/s│
│   YaS    │  6.7 GB/s ±  10.8 MB/s│  6.3 GB/s ±  23.2 MB/s│  7.6 GB/s ±   7.2 MB/s│
│ CmwLight │  8.7 GB/s ±  18.3 MB/s│  8.2 GB/s ±  23.8 MB/s│  8.2 GB/s ±  13.4 MB/s│
└──────────┴───────────────────────┴───────────────────────┴───────────────────────┘
┌─protocol─┬────────ALWAYS─────────┬────────LENIENT────────┬────────IGNORE─────────┐
│   YAML   │101.9 MB/s ± 800.6 kB/s│ 99.8 MB/s ±   1.0 MB/s│101.0 MB/s ± 441.6 kB/s│
│   Json   │329.0 MB/s ±   1.1 MB/s│326.5 MB/s ±   1.8 MB/s│327.5 MB/s ± 988.7 kB/s│
│   YaS    │  6.6 GB/s ±  12.3 MB/s│  6.3 GB/s ±   4.6 MB/s│  7.6 GB/s ±  21.3 MB/s│
│ CmwLight │  8.6 GB/s ±  14.3 MB/s│  8.1 GB/s ± 162.4 MB/s│  8.5 GB/s ±  88.8 MB/s│
└──────────┴───────────────────────┴───────────────────────┴───────────────────────┘
┌─protocol─┬────────ALWAYS─────────┬────────LENIENT────────┬────────IGNORE─────────┐
│   YAML   │101.3 MB/s ± 258.9 kB/s│ 98.7 MB/s ±  75.9 kB/s│100.6 MB/s ± 133.2 kB/s│
│   Json   │325.7 MB/s ±   1.9 MB/s│329.8 MB/s ± 682.1 kB/s│330.4 MB/s ± 803.4 kB/s│
│   YaS    │  6.7 GB/s ±   7.9 MB/s│  6.3 GB/s ±   6.4 MB/s│  7.6 GB/s ±   8.7 MB/s│
│ CmwLight │  8.7 GB/s ±  16.6 MB/s│  8.1 GB/s ±  89.6 MB/s│  8.4 GB/s ±  54.7 MB/s│
└──────────┴───────────────────────┴───────────────────────┴───────────────────────┘
```
Not bad, but I wonder whether this could be pushed up to 8...9 GB/s for both YaS and CmwLight... :thinking:

src/serialiser/include/opencmw.hpp
```diff
@@ -48,6 +48,8 @@ inline constexpr bool isReflectableClass() {
     }
     return false;
 }
+template<typename T>
+constexpr bool isReflectableClass_v = isReflectableClass<T>();
```
RalphSteinhagen, line 52: Why the new/additional definition?

Do we want to change `isReflectableClass<T>)` from a constexpr function declaration to templated constexpr boolean expression `isReflectableClass_v<T>`?  I wouldn't mind as long as it is consistent and we have only one definition...

To note: maybe we should later revisit/stream-line other similar expression too..? :thinking:

RalphSteinhagen (approved):
Good work on adding nested fields to the CmwLight protocol (de-)serialisers! :+1:

The performance dropped from (temporarily) 9.2 GB/s to now ~8.4 GB/s... not too bad. 
Maybe we (at a future point-of-time) could revisit and stream-line the IoSerialiser with the info/experience gained.

Some optimization ideas:
  * drop computing the field hash function (YaS)
  * introduce the field count for nested objects for all applicable protocols -- N.B. should be faster to for-loop (if the number is known) than to read field-by-field until the end-marker
  * other algorithmic optimisation gains.
  In any case, this would need to be synchronised to the java implementation.

PR is good and could be merged as-is.

src/serialiser/include/opencmw.hpp
wirew0rm: sorry, I just saw this after merging. I preferred not to have the additional parentheses after the template argument as they are hard to write and read for complex/nested template arguments(The case where it bothered me originally has been refactored into something simpler in the meantime). I don't know if it is possible to remove the constexpr function, my line of thinking was to have this as a shorthand similar to the `A<T>::value` case, just that it gets rid of the parantheses instead of the static member access. If it works directly, I think I would prefer to just have the expression template, but I'm not sure I'm seeing all the consequences this has.

PR #112, opened 2022-01-18
initial YAML serialiser implementation

RalphSteinhagen:
N.B. this is primarily intended for exchange of config-data files not large volume measurement data


src/serialiser/include/IoSerialiserYAML.hpp
```diff
@@ -0,0 +1,507 @@
...
+                break;
+            }
+            [[maybe_unused]] const auto n = moveToEndOfLine(buffer); // skip commented or empty lines
+        }
+        result.headerStart     = buffer.position();
+        const auto indentation = consumeWhitespace(buffer);
+        if (indentation % IOSERIALISER_YAML_INDENTATION) {
+            throw ProtocolException("YAML: non canonical indentation depth {} at buffer position {}", indentation, buffer.position());
+        }
+        const auto newHierarchyDepth = static_cast<uint8_t>(indentation / IOSERIALISER_YAML_INDENTATION);
```
wirew0rm, line 473: To really parse all valid YAML we would have to allow arbitrary indentation depths, which can be different on different levels/parts of the file. For simplicity, I think it is ok to make this assumption but we should at least document it somewhere. From an end user perspective, of course everything that does behave in a nonstandard way is not really nice, so if there was a simple way to use `IOSERIALISER_YAML_INDENTATION` only for the serialiser, i would prefer that. Wouldn't have to be on the first iteration of course.

src/serialiser/include/IoSerialiserYAML.hpp
```diff
@@ -0,0 +1,507 @@
...
+            if constexpr (is_annotated<DataType>) {
+                const auto unit        = rawValue.getUnit().empty() ? "" : fmt::format(" - [{}]", rawValue.getUnit());
+                const auto description = rawValue.getDescription().empty() ? "" : fmt::format(" - {}", rawValue.getDescription());
+                buffer.put<opencmw::IoBuffer::MetaInfo::WITHOUT>(fmt::format("{}{:<{}}# {}\t{}{}\n", rawFormat, ' ', spaceUntilMeta, typeName<StrippedDataType>, unit, description));
+                return;
+            }
+            buffer.put<opencmw::IoBuffer::MetaInfo::WITHOUT>(fmt::format("{}{:<{}}# {}\n", rawFormat, ' ', spaceUntilMeta, typeName<StrippedDataType>));
+            return;
+        }
+        // need line-break to not collide with the meta information
```
wirew0rm, line 406: If the field name is so long(or there are just many indentation levels), that it exceeds `IOSERIALISER_YAML_META_POSTION`, the serialiser crashes. Probably an edge case which we don't need to support for our usecase, but it should report an error instead of crashing.
But I really like how it automatically switches to multiline if it would collide :+1:

src/serialiser/test/IoSerialiserYAML_tests.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+using opencmw::Annotated;
+using opencmw::NoUnit;
+using opencmw::ProtocolCheck;
+using opencmw::ExternalModifier::RO;
+using opencmw::ExternalModifier::RW;
+using opencmw::ExternalModifier::RW_DEPRECATED;
+using opencmw::ExternalModifier::RW_PRIVATE;
+
+struct NestedDataY {
```
wirew0rm, line 28: Had to change the name here for it to be unique so I could add bool members without confusing refl-cpp

wirew0rm (approved):
Very nice! I found some small things when testing which I changed on the spot (see last commit) and added some comments for things we might want to consider adding or changing later.
Generally in terms of unit testing, I think we should also add some tests that test against static testcases or reference implementations for all serialisers as a long-medium term goal.
Overall this looks very good! :+1:

concepts/serialiser/IoSerialiserBenchmark.cpp
```diff
@@ -0,0 +1,85 @@
...
+ * benchmark exemplary output:
+ * [..]
+ * IO Serializer (POCO, YaS, 0) throughput = 7.5 GB/s for 29.6 kB per test run (took 395.8 ms)
+ * IO Serializer (POCO, YaS, 0) throughput = 7.5 GB/s for 29.6 kB per test run (took 397.1 ms)
+ * IO Serializer (POCO, YaS, 0) throughput = 7.5 GB/s for 29.6 kB per test run (took 395.8 ms)
+ * ┌─protocol─┬────────ALWAYS─────────┬────────LENIENT────────┬────────IGNORE─────────┐
+ * │   YAML   │ 95.7 MB/s ±   2.9 MB/s│ 99.4 MB/s ± 232.1 kB/s│ 86.5 MB/s ± 190.7 kB/s│
+ * │   Json   │334.9 MB/s ± 780.6 kB/s│333.4 MB/s ± 560.1 kB/s│335.4 MB/s ± 822.2 kB/s│
+ * │   YaS    │  6.9 GB/s ±   6.2 MB/s│  6.4 GB/s ±  73.3 MB/s│  7.5 GB/s ±  16.4 MB/s│
+ * └──────────┴───────────────────────┴───────────────────────┴───────────────────────┘
```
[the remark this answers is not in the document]
RalphSteinhagen: done.

src/serialiser/include/IoSerialiserYAML.hpp
RalphSteinhagen: fixed and done

PR #119, opened 2022-01-25
IoSerialiserJson: implement Map<String,V> serialisation and small fixes

wirew0rm:
Implement a serialiser for maps with strings as keys.
Removes the trailing commas after the last field of an object.
Implements the deserialiser for the MultiArray type (was missing until now and related to the map deserialiser).


ivan-cukic (approved):
Looks good to me

PR #121, opened 2022-01-27
Initialise uninitialised variables in Mustache tests

ivan-cukic:


RalphSteinhagen (approved):
Thanks for the quick fix!

PR #125, opened 2022-02-04
TimingCtx: re-enabled serialising of 'bpcts' field

RalphSteinhagen:
... and made both selector and bpcts Annotated<...> fields. Should yield better self-documentation and type checks.

### DIFF
--- a/src/core/include/TimingCtx.hpp
+++ b/src/core/include/TimingCtx.hpp
@@ -1,8 +1,6 @@
 #ifndef OPENCMW_CORE_TIMINGCTX_H
 #define OPENCMW_CORE_TIMINGCTX_H
 
-#include <fmt/format.h>
-
 #include <algorithm>
 #include <charconv>
 #include <chrono>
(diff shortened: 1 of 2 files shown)


src/core/include/TimingCtx.hpp
```diff
@@ -11,13 +9,20 @@
...
+#include <units/chrono.h>
+#include <units/isq/dimensions/time.h>
+#include <units/isq/si/time.h>
+
 namespace opencmw {
 
 namespace detail {
-constexpr uint32_t const_hash(char const *input) noexcept { return *input ? static_cast<uint32_t>(*input) + 33 * const_hash(input + 1) : 5381; } // NOLINT
+constexpr uint32_t           const_hash(char const *input) noexcept { return *input ? static_cast<uint32_t>(*input) + 33 * const_hash(input + 1) : 5381; } // NOLINT
+constexpr basic_fixed_string TIMING_DESCRIPTION = "FAIR Timing Selector (e.g. 'FAIR.SELECTOR.C=1:S=2:P=3:T=4'. 'FAIR.SELECTOR.ALL' or 'ALL'";
```
wirew0rm, line 21: Would be nice to have this close to the field itself, but I guess that is difficult without polluting the TimingCtx object.

wirew0rm (approved):
Looks good!

src/core/include/TimingCtx.hpp
RalphSteinhagen: The choice is either having it `constexpr` (outside the class definition) or `const static` (inside the class definition). 

The rationale behind using `constexpr` was/is that this is easier to inline/optimised away by the compiler.

PR #137, opened 2022-03-02
Fix json serialisation of nested objects

wirew0rm:
This fixes a serialisation error discovered by @frankosterfeld. Nested objects which where not at the end of the class would lead to malformed json because the comma after the object was not added.


RalphSteinhagen (approved):
LGTM! Nice reproduce-fix commits! :+1:

PR #142, opened 2022-03-03
Cmake fetch content

wirew0rm:
Attempt to get rid of the dependency on conan.
All direct dependencies of opencmw behave more or less nice with cmake's fetchContent.
This means that all dependencies are imported as cmake subprojects, making their targets available directly from opencmw.
The indirect dependencies are more problematic, as they are often not cmake based and difficult to build in tree.

direct dependencies:
- refl-cpp
- mp-units
- fmt
- cpp-httplib
- libzmq

indirect dependencies
- openssl (via httplib, optional, no cmake)
- zlib, brotli (via httplib, optional for compressed requests)
- gsl-lite (via mp-units, header only, cmake supported)

Graph of the all targets in opencmw and its dependencies
image

TODO:
- [x] allow opencmw and its dependencies to be compiled without conan
- [x] Enable caching for the ci
- [x] cleanup
- [ ] add a way to allow using system libraries (mostly only useful for zeromq, as the others are all header only)
- [x] disable tests, benchmarks and examples of dependencies
- [ ] figure out which optional dependencies are needed and how to add them
  - for now the dependencies are just resolved from the system if available using `find_package`
- [ ] fix the install directives, find out whether and where to install dependencies
- [x] make opencmw consumable as a sub-project
- [x] support for other clang and CI workflow for it
  - [ ] later: other compilers, OSes, WASM
- [ ] static linking?
- [ ] reenable -Werror (commented out)
- [ ] disable gsl-lite warnings and bring `SYSTEM` includepath upstream in rxcpp


.github/workflows/build_cmake.yml
```diff
@@ -48,7 +44,7 @@ jobs:
         cache-name: cache-fetchContent-cache
       with:
         path: ${{runner.workspace}}/build/_deps
```
RalphSteinhagen, line 46: very smart caching strategy! :+1:

RalphSteinhagen (approved):
Very nice addition, simplification and reduction of external dependencies. :+1:

Additional open pertinent points should be transferred to issues eventually so that they are not being lost.

PR #147, opened 2022-03-11
Fix DNS handling

frankosterfeld:
The forwarding of service names to another broker (via "dnsAddress")
did not work at all. Also clean up the DNS entries and make their
order predictable, for better testability.

Improve the test to cover more of the DNS logic.


wirew0rm (approved):
Very nice, thanks for all the work.
On reading i found some of the code a bit hard to read especially when there where multiple levels of nested Pair types. I added some small changes which should make it a bit easier to read for people not intimately familiar with the codebase.

PR #150, opened 2022-03-24
Suppress compiler warnings for dependent header only libraries

wirew0rm:
Suppress compiler warnings for header only libraries by wrapping their include in with a `#pragma diagnostic push/pop`.  The cause of these warnings cannot be fixed on our side, so we have to wait for upstream to fix them. This is more verbose than using the cmake `SYSTEM` keyword to add them to `-Isystem` which can hide real bugs and has other side effects. Since `gsl-lite` is a dependency of `mp-units`, the suppression has to be added around every header inclusion of mp-units.


RalphSteinhagen (approved):
LGTM! :+1:

PR #152, opened 2022-03-30
Improve build/example section of README.md

alexxcons:
- Simple build instruction, with reference to more advanced options
- Make clear that building opencmw-cpp itself is not required when the
intention is, to create a new service
- Reference to opencmw-cpp-example instead of Java example

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,23 +90,7 @@ Note that building from source is only required if you want to modify opencmw-cp
 If you only want to make use of opencmw-cpp in your project to implement a service, it is not required to clone/build opencmw-cpp.
 In that case, rather take a look at the project [opencmw-cpp-example](https://github.com/alexxcons/opencmw-cpp-example), which demonstrates how a simple, first service can be implemented.
 
-
-Building on linux operating systems:
-
-```
-git clone https://github.com/alexxcons/opencmw-cpp.git
-cd opencmw-cpp
-mkdir build
-ccmake -S . -B ./build
-```
-* press 'c' and wait until 'configure' is done, check for errors and try to install missing dependencies
-* If there are no errors, press 'g' and wait until 'generate' is done and start the build:
-
-```
-cmake --build ./build
-```
-
-For detailed, advanced options or installation on non-linux operating systems, please check the [detailed build instructions](docs/BuildInstructions.md).
+For concrete build instructions, please check the [build instructions page](docs/BuildInstructions.md).
 
 ### Example
 
--- a/README.md
+++ b/README.md
@@ -83,61 +83,36 @@ provides also an optional light-weight `constexpr` annotation template wrapper `
 that in turn can be used to (re-)generate and document the class definition (e.g. for other programming languages or projects that do not have the primary domain-object definition at hand) 
 or to generate a generic [OpenAPI](https://swagger.io/specification/) definition. More details can be found [here](docs/CompileTimeSerialiser.md).
 
+### Building from source
 
-### Example
-The following provides some flavour of how a simple service can be implemented using OpenCMW with only a few lines of
-custom user-code ([full sample](https://github.com/fair-acc/opencmw-java/tree/createReadme/server-rest/src/test/java/io/opencmw/server/rest/samples/BasicSample.java)):
+Note that building from source is only required if you want to modify opencmw-cpp itself.
 
-```Java
-// TODO: Java Concept to be ported/implemented in C++
-@MetaInfo(description = "My first 'Hello World!' Service")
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -83,61 +83,36 @@ provides also an optional light-weight `constexpr` annotation template wrapper `
 that in turn can be used to (re-)generate and document the class definition (e.g. for other programming languages or projects that do not have the primary domain-object definition at hand) 
 or to generate a generic [OpenAPI](https://swagger.io/specification/) definition. More details can be found [here](docs/CompileTimeSerialiser.md).
 
+### Building from source
 
-### Example
-The following provides some flavour of how a simple service can be implemented using OpenCMW with only a few lines of
```
RalphSteinhagen, line 88: This code example example is important and provides a flavour of the library and concepts involved.

While the syntax is different ins some parts, we want to keep the examples/documentation pretty much aligned between the C++ and Java version of OpenCMW.

Ideally, using language specific code tabs like:
image

While GH markdown does not support this out-of-the box (N.B. to be further investigated), we should setup HTML doc  page where this is done, and at least keep/leave the corresponding C++ version here.

The code syntax flavour is more important than how to use `cmake` (which pretty standard anyway).

README.md
```diff
@@ -83,61 +83,36 @@ provides also an optional light-weight `constexpr` annotation template wrapper `
...
+Note that building from source is only required if you want to modify opencmw-cpp itself.
+
+If you only want to make use of opencmw-cpp in your project to implement a service, it is not required to clone/build opencmw-cpp.
+In that case, rather take a look at the project [opencmw-cpp-example](https://github.com/alexxcons/opencmw-cpp-example), which demonstrates how a simple, first service can be implemented.
+
+
+Building on linux operating systems:
+
+```
+git clone https://github.com/alexxcons/opencmw-cpp.git
```
RalphSteinhagen, line 97: wrong github repo

README.md
```diff
@@ -83,61 +83,36 @@ provides also an optional light-weight `constexpr` annotation template wrapper `
...
+cd opencmw-cpp
+mkdir build
+ccmake -S . -B ./build
```
RalphSteinhagen, line 100: Aside being unnecessarily complex for uninitiated cmake users, we should not use UI-style and in-source build flavour cmake.

A simpler and more intuitive way which is also copy-paste safe would be:
```
cd opencmw-cpp
mkdir build; cd build
cmake .. -DCMAKE_BUILD_TYPE=Release # for debug builds use 'Debug'
make -j 4
```
...Execute tests via `ctest` in `.` and samples in `./concepts/<core|cmrc|disruptor|majordomo|serialiser>/`...

Ideally this would be also wrapped in a by-language code tab to keep the doc in sync.

README.md
```diff
@@ -83,61 +83,36 @@ provides also an optional light-weight `constexpr` annotation template wrapper `
...
 ```
+* press 'c' and wait until 'configure' is done, check for errors and try to install missing dependencies
+* If there are no errors, press 'g' and wait until 'generate' is done and start the build:
+
+```
+cmake --build ./build
```
RalphSteinhagen, line 106: ... can cause confusion and needless complex cmd-line syntax

README.md
```diff
@@ -83,61 +83,36 @@ provides also an optional light-weight `constexpr` annotation template wrapper `
...
+```
+
+For detailed, advanced options or installation on non-linux operating systems, please check the [detailed build instructions](docs/BuildInstructions.md).
```
RalphSteinhagen, line 109: is probably the better place to add these cmake snippets

RalphSteinhagen (changes requested):

README.md
RalphSteinhagen: To give a specific flavour of documentation style we want to emulate have a look at, for example: https://javalin.io/

README.md
alexxcons: Alright .. I just tried to use what was suggested in `docs/BuildInstructions.md`.

README.md
alexxcons: Ok, will move all the snippets to `docs/BuildInstructions.md` and only leave a link here.

alexxcons:
Fixed now, please re-check.

- Moved all build instruction snippets into BuildInstructions.md
- I did not find a way to use language specific code tabs in github, so that will be for later.
- I just kept the current java example as it is for now. If I find time, I will try to add the same example in C++

RalphSteinhagen (approved):
Thanks for the changes! LGTM! :+1:

PR #156, opened 2022-05-03
Transactional Settings w/o and w/ FAIR TimingCtx-Multiplexing

RalphSteinhagen:
A transactional setting that can be used to store a value of the user-defined type T for a user-defined transaction token, a given history length, and (optional) FAIR TimingCtx.

API Examples:
```cpp
opencmw::SettingBase<int,int, std::string, 16, std::chrono::seconds, 3600 * 24, 10> settings;

auto [ok1, timeStamp1] = a.commit(42);  // store 42 in settings
auto [ok2, timeStamp2] = a.commit(43);  // store 43 in settings

assert(settings.get() == 43);
assert(a.get().settingValue == 43);     // get the latest value
assert(a.get() == 43);                  // get the latest value --short-hand notation
assert(settings.get(timeStamp2) == 43); // get the first valid value since timeStamp2
assert(settings.get(-1) == 42);         // get the second to last value
assert(settings.get(timeStamp1) == 42); // get the first valid value since timeStamp2

auto [ok3, timeStamp3] = a.stage(53, "transactionToken#1"); // stage 53 in settings, N.B. setting is not yet committed
assert(settings.get() == 43);
auto [ok4, timeStamp4] = a.commit("transactionToken#1"); // commit transaction
assert(settings.get() == 53);
```

The `SettingCtx` behaves similar but has additionally multiplexing capabilities:
```cpp
opencmw::CtxSetting<int, std::string, 16, std::chrono::seconds, 3600 * 24, 10> settings; // history length of 24 hours, time-out of 10 seconds

assert(settings.commit(TimingCtx(1, 1, 1, 1), 101)); // store 101 for 'TimingCtx(cid=1, sid=1, pid=1, gid=1)'
assert(settings.commit(TimingCtx(1, 1, 1), 102));    // store 102 for 'TimingCtx(cid=1, sid=1, pid=1, gid=-1)'
assert(settings.commit(TimingCtx(1, 1), 103));       // store 103 for 'TimingCtx(cid=1, sid=1, pid=-1, gid=-1)'
assert(settings.commit(TimingCtx(1), 104));          // store 104 for 'TimingCtx(cid=1, sid=-1, pid=-1, gid=-1)'

// retrieve settings
assert(settings.get(TimingCtx(1)) == 104);
assert(settings.get(TimingCtx(1, 1, 1, 1)) == 101);
assert(settings.get(TimingCtx(1, 1, 1, 1)).timingCtx == TimingCtx(1, 1, 1, 1));
assert(settings.get(TimingCtx(1, 1, 1, 1)) == TimingCtx(1, 1, 1, 1)); // short-hand notation
assert(settings.get(TimingCtx(1, 1, 1, 2)) == 102); // no setting for 'gid=2' -> fall back to 'gid=-1'
assert(settings.get(TimingCtx(1, 1, 1, 2)) == TimingCtx(1, 1, 1));
assert(settings.get(TimingCtx(1, 1, 2, 2)) == 103); // no setting for 'pid=2' and 'gid=2' -> fall back to 'pid=gid=-1'
assert(settings.get(TimingCtx(1, 1, 2, 2)) == TimingCtx(1, 1));

// stage new settings
auto [committed, commitTimeStamp] = settings.commit(TimingCtx(1), 55); // immediately store 55 for 'TimingCtx(cid=1)'
assert(not settings.stage(TimingCtx(1), 555, "tokenNo1"));
assert(not settings.stage(TimingCtx(1), 556, "tokenNo2"));

assert(settings.get(TimingCtx(1)) == 55); // staged settings not yet commited
assert(settings.get(TimingCtx(2)) != TimingCtx(2));
assert(settings.commit("tokenNo1"));
assert(settings.get(TimingCtx(1)) == 555);
assert(settings.commit("tokenNo2"));
assert(settings.get(TimingCtx(1)) == 556);
```

This is an API prototype looking for user-feedback ... thanks in advance.

### DIFF
--- a/src/core/include/TimingCtx.hpp
+++ b/src/core/include/TimingCtx.hpp
@@ -5,6 +5,7 @@
 #include <charconv>
 #include <chrono>
 #include <exception>
+#include <functional>
 #include <opencmw.hpp>
 #include <ranges>
 #include <string_view>
@@ -20,6 +21,11 @@
 namespace opencmw {
 
 namespace detail {
+template<class T>
+inline constexpr void hash_combine(std::size_t &seed, const T &v) noexcept {
+    std::hash<T> hasher;
+    seed ^= hasher(v) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
+}
 constexpr uint32_t           const_hash(char const *input) noexcept { return *input ? static_cast<uint32_t>(*input) + 33 * const_hash(input + 1) : 5381; } // NOLINT
 constexpr basic_fixed_string TIMING_DESCRIPTION = "FAIR Timing Selector (e.g. 'FAIR.SELECTOR.C=1:S=2:P=3:T=4'. 'FAIR.SELECTOR.ALL' or 'ALL'";
 } // namespace detail
@@ -87,6 +93,17 @@ public:
         if (!isWildcard(_gid)) { segments.emplace_back(fmt::format("T={}", _gid)); }
         // clang-format on
         return fmt::format("{}{}", SELECTOR_PREFIX, fmt::join(segments, ":"));
+    }
+
+    [[nodiscard]] std::size_t hash() const noexcept {
+        parse();
+        std::size_t seed = 0;
+        detail::hash_combine(seed, _cid);
+        detail::hash_combine(seed, _sid);
+        detail::hash_combine(seed, _pid);
+        detail::hash_combine(seed, _gid);
+        detail::hash_combine(seed, bpcts.value());
+        return seed;
     }
 
     void parse() const {
@@ -186,14 +203,25 @@ private:
     }
 };
 
-[[nodiscard]] inline bool operator==(const TimingCtx &lhs, const std::string_view &rhs) { return (lhs.bpcts == 0) && (lhs.selector.value() == rhs); }
+inline static const TimingCtx NullTimingCtx = TimingCtx{};
 
-inline std::ostream      &operator<<(std::ostream &os, const opencmw::TimingCtx &v) {
+[[nodiscard]] inline bool     operator==(const TimingCtx &lhs, const std::string_view &rhs) { return (lhs.bpcts == 0) && (lhs.selector.value() == rhs); }
+
+inline std::ostream          &operator<<(std::ostream &os, const opencmw::TimingCtx &v) {
     return os << fmt::format("{}", v);
 }
 
 } // namespace opencmw
 ENABLE_REFLECTION_FOR(opencmw::TimingCtx, selector, bpcts);
+
+namespace std {
+template<>
+struct hash<opencmw::TimingCtx> {
+    [[nodiscard]] size_t operator()(const opencmw::TimingCtx &ctx) const noexcept {
+        return ctx.hash();
+    }
+};
+} // namespace std
 
 template<>
 struct fmt::formatter<opencmw::TimingCtx> {
--- a/src/core/test/TimingCtx_tests.cpp
+++ b/src/core/test/TimingCtx_tests.cpp
@@ -13,6 +13,14 @@ TEST_CASE("Basic TimingCtx tests", "[TimingCtx][basic]") {
     REQUIRE("ALL" == TimingCtx("ALL"));
     REQUIRE(TimingCtx("all") == "ALL");
     REQUIRE(TimingCtx("ALL").bpcts.value() == 0);
+    REQUIRE(TimingCtx("All").hash() != 0);
+    REQUIRE(TimingCtx("All").hash() == TimingCtx("FAIR.SELECTOR.ALL").hash());
+    REQUIRE(TimingCtx("All").hash() == TimingCtx("FAIR.SELECTOR.C=-1").hash());
+    REQUIRE(TimingCtx("All").hash() == TimingCtx("FAIR.SELECTOR.C=-1:S=-1").hash());
+    REQUIRE(TimingCtx("All").hash() == TimingCtx("FAIR.SELECTOR.C=-1:S=-1:P=-1").hash());
+    REQUIRE(TimingCtx("All").hash() == TimingCtx("FAIR.SELECTOR.C=-1:S=-1:P=-1:T=-1").hash());
+    REQUIRE(TimingCtx("All").hash() != TimingCtx("FAIR.SELECTOR.C=-1:S=-1:P=-1:T=0").hash());
+
     auto changeMyFields = TimingCtx("ALL");
     REQUIRE(changeMyFields == "ALL");
     REQUIRE(changeMyFields.cid() == -1);
--- a/src/disruptor/include/Transactions.hpp
+++ b/src/disruptor/include/Transactions.hpp
@@ -1,0 +1,447 @@
+#ifndef OPENCMW_CPP_TRANSACTIONS_HPP
+#define OPENCMW_CPP_TRANSACTIONS_HPP
+
+#include <atomic>
+#include <cassert>
+#include <chrono>
+#include <concepts>
+#include <functional>
+#include <list>
+#include <unordered_map>
+#include <utility>
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-nonliteral"
+#include <fmt/chrono.h>
+#pragma GCC diagnostic pop
+
+#include <disruptor/RingBuffer.hpp>
+#include <opencmw.hpp>
+#include <ReaderWriterLock.hpp>
+#include <TimingCtx.hpp>
+
+namespace opencmw {
+
+namespace settings {
+template<typename T>
+struct alignas(disruptor::kCacheLine) node {
+    using TimeStamp               = std::chrono::system_clock::time_point;
+    std::shared_ptr<T> value      = std::make_shared<T>();
+    TimeStamp          validSince = std::chrono::system_clock::now();
+    mutable TimeStamp  lastAccess = std::chrono::system_clock::now();
+    node()                        = default;
+    explicit node(T &&in)
+        : value(std::make_shared<T>(FWD(in))){};
+    constexpr void            touch() const noexcept { lastAccess = std::chrono::system_clock::now(); }
+    explicit(false) constexpr operator T const &() const noexcept { return *value; }
+};
+static_assert(sizeof(node<int>) % disruptor::kCacheLine == 0, "node size must be cache line aligned");
+
+struct TransactionResult {
+    using TimeStamp = std::chrono::system_clock::time_point;
+    const bool                isCommitted;
+    const TimeStamp           timeStamp;
+
+    explicit(false) constexpr operator bool const &() const noexcept { return isCommitted; }
+    explicit(false) constexpr operator TimeStamp const &() const noexcept { return timeStamp; }
+    auto                      operator<=>(const TransactionResult &) const noexcept = default;
+};
+
+template<typename T>
+struct CtxResult {
+    const TimingCtx           timingCtx;
+    const settings::node<T>  &settingValue;
+
+    explicit(false) constexpr operator TimingCtx const &() const noexcept { return timingCtx; }
+    explicit(false) constexpr operator T const &() const noexcept { return settingValue; }
+    auto                      operator<=>(const CtxResult &) const noexcept = default;
+};
+
+} // namespace settings
+
+template<std::equality_comparable TransactionToken>
+inline static const TransactionToken NullToken = TransactionToken{};
+
+/**
+ * @brief A thread-safe settings wrapper that supports multiple stage/commit transactions and history functionality.
+ *
+ * Example:
+ * @code
+ * opencmw::SettingBase<int,int, std::string, 16, std::chrono::seconds, 3600 * 24, 10> settings;
+ *
+ * auto [ok1, timeStamp1] = settings.commit(42); // store 42 in settings
+ * auto [ok2, timeStamp2] = settings.commit(43); // store 43 in settings
+ *
+ * assert(settings.get() == 43); // get the latest value
+ * assert(settings.get(timeStamp2) == 43); // get the first isCommitted value since timeStamp2
+ * assert(settings.get(-1) == 42); // get the second to last value (for lib development only)
+ * assert(settings.get(timeStamp1) == 42); // get the first isCommitted value since timeStamp1
+ *
+ * auto [ok3, timeStamp3] = settings.stage(53, "transactionToken#1"); // stage 53 in settings, N.B. setting is not yet committed
+ * assert(settings.get() == 43); // get the latest value
+ * auto [ok4, timeStamp4] = settings.commit("transactionToken#1"); // commit transaction
+ * assert(settings.get() == 53); // get the latest value
+ * @endcode
+ *
+ * @tparam T is the user-supplied setting type, for simple settings U and T are identical (see CtxSettings for an example where it isn't).
+ * @tparam U is the internally stored setting type that may include additional meta data.
+ * @tparam TransactionToken unique identifier with which to store/commit transactions.
+ * @tparam N_HISTORY the maximum number of setting history
+ * @tparam TimeDiff the std::chrono::duration time-base for the time-outs
+ * @tparam timeOut maximum time given in units of TimeDiff after which a setting automatically expires if unused. (default: -1 -> disabled)
+ * @tparam timeOutTransactions maximum time given in units of TimeDiff after which a transaction automatically expires if not being committed. (default: -1 -> disabled)
+ */
+template<std::movable T, std::movable U = T, std::equality_comparable TransactionToken = std::string, std::size_t N_HISTORY = 1024, typename TimeDiff = std::chrono::seconds, int timeOut = -1, int timeOutTransactions = -1>
+requires(opencmw::is_power2_v<N_HISTORY> &&N_HISTORY > 8) class SettingBase {
+    using TimeStamp  = std::chrono::system_clock::time_point;
+    using RingBuffer = opencmw::disruptor::RingBuffer<settings::node<U>, N_HISTORY, disruptor::BusySpinWaitStrategy, disruptor::MultiThreadedStrategy>;
+    using Sequence   = disruptor::Sequence;
+    //
+    constexpr static std::size_t BUFFER_MARGIN                             = 8;
+    alignas(disruptor::kCacheLine) std::shared_ptr<RingBuffer> _ringBuffer = std::make_shared<RingBuffer>();
+    alignas(disruptor::kCacheLine) std::shared_ptr<Sequence> _sequenceHead = std::make_shared<Sequence>();
+    alignas(disruptor::kCacheLine) std::shared_ptr<Sequence> _sequenceTail = std::make_shared<Sequence>(0);
+    alignas(disruptor::kCacheLine) mutable ReaderWriterLock _historyLock{};
+    alignas(disruptor::kCacheLine) mutable std::atomic<bool> _transactionListLock{ false };
+    std::list<std::pair<TransactionToken, settings::node<T>>> _transactionList;
+    alignas(disruptor::kCacheLine) const std::function<U(const U &, T &&)> _onCommit;
+
+public:
+    using Node              = settings::node<U>;
+    using TransactionResult = settings::TransactionResult;
+    SettingBase()
+        : SettingBase([](const U & /*old*/, T &&in) -> U { return static_cast<U>(std::move(in)); }){};
+    template<class Fn>
+    requires std::is_invocable_r_v<U, Fn &&, const U &, T &&>
+    explicit SettingBase(Fn &&onCommit)
+        : _onCommit(FWD(onCommit)) {
+        _ringBuffer->tryPublishEvent([](Node &&eventData, std::int64_t) { eventData.value = std::make_shared<U>(U()); }); // init with default setting
+        _sequenceHead->setValue(_sequenceHead->value());
+        _ringBuffer->addGatingSequences(std::vector{ _sequenceHead, _sequenceTail });
+    }
+    [[nodiscard]] std::size_t nHistory() const noexcept {
+        _historyLock.scopedGuard<ReaderWriterLockType::READ>();
+        return static_cast<std::size_t>(_sequenceHead->value() - _sequenceTail->value() + 1);
+    }
+    ReaderWriterLock &historyLock() noexcept { return _historyLock; }
+
+    TransactionResult stage(T &&t, const TransactionToken &transactionToken = NullToken<TransactionToken>, const TimeStamp &now = std::chrono::system_clock::now()) {
+        if (transactionToken.empty()) {
+            const auto isCommitted = _ringBuffer->tryPublishEvent([&t, this, &now](Node &&eventData, std::int64_t sequence) {
+                const auto oldValue  = get();
+                eventData.value      = std::make_shared<U>(_onCommit(*oldValue.value, FWD(t)));
+                eventData.validSince = now;
+                eventData.lastAccess = now;
+                _sequenceHead->setValue(sequence);
+            });
+            retireExpired();
+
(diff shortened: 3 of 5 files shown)


alexxcons:
The usage of git syntax / staging is nice !

Maybe not related to this MR (sorry, I still did not find the time to properly test opencmw) :

```
assert(settings.get() == 43); // get the latest value
assert(settings.get(timeStamp2) == 43); // get the first valid value since timeStamp2
assert(settings.get(-1) == 42); // get the second to last value
assert(settings.get(timeStamp1) == 42); // get the first valid value since timeStamp2

```
The need to add a comment after each of these calls arises from the rather simplicic methos name. IMO methods names should reflect what they are supposed to do. Using the same name might look nice in the API, but for usage, ans specially for readability of code it has disadvantages. If the name of the methods would be `get_last()`, `get_last_since(timestamp)` and `get_oldest()`, the comments would not be required. Specially `get(-1)` is real bad. If I would see that without comment, I would have no idea what is going on, and would need to check the documentation of the method.

alexxcons:
> assert(settings.get(TimingCtx(1)).second == 55); // get the last committed value for 'TimingCtx(cid=1, sid=1, pid=-1, gid=-1)'

I would expect to get the latest commit for cid=1 by this call, no matter what sid/pid/gid   .. though maybe teh comment is just unprecise here.

alexxcons:
```
assert(settings.commit(TimingCtx(1, 1, 1, 1), 101).first); // store 101 in settings for 'TimingCtx(cid=1, sid=1, pid=1, gid=1)'
...
assert(settings.get(TimingCtx(1, 1, 1, 1)).second == 101);
```
Using .first / .second here (probably a `std::pair` ) is not very readable. Imo it would be better to use a structure and give proper names to the items. E.g. `value` and `context`.

alexxcons:
> assert(settings.get(TimingCtx(1, 1, 1, 2)).second == 102); // no setting for 'gid=2' exists yet -> fall back to 'gid=-1'

Imo such a silent fallback is rather dangerous and could result in setting wrong values to devices. Imo it would make more sense to dont do anything, for a device, and possibly signal a software interlock if a needed setting is missing. Though maybe I overlook something.  How can I actually figure out that there is no setting provided for a specific gid ?

alexxcons:
I think the stage/commit API is good, and I suppose would be sufiicient to cover transactional settings. Though I dont have the edge-cases for transactional settings in mind. It would be good if as well Dominic Day would review this PR, since he is taking care for the transactional settings in Fesa.

Some more questions:
- What happens if I stage twice for the same token ? I would expect an overwrite of the first staged value.
- What happens with data which is staged, but never commit ? I would expect that it just cycle out if the buffer rolls over.

RalphSteinhagen:
>

RalphSteinhagen:
> ```
> assert(settings.commit(TimingCtx(1, 1, 1, 1), 101).first); // store 101 in settings for 'TimingCtx(cid=1, sid=1, pid=1, gid=1)'
> ...
> assert(settings.get(TimingCtx(1, 1, 1, 1)).second == 101);
> ```
> 
> Using .first / .second here (probably a `std::pair` ) is not very readable. Imo it would be better to use a structure and give proper names to the items. E.g. `value` and `context`.

Understood. The idea was to use structured binding. In most cases one would use only one of the return values... the others are more expert-type fields.

RalphSteinhagen:
> > assert(settings.get(TimingCtx(1)).second == 55); // get the last committed value for 'TimingCtx(cid=1, sid=1, pid=-1, gid=-1)'
> 
> I would expect to get the latest commit for cid=1 by this call, no matter what sid/pid/gid .. though maybe the comment is just unprecise here ?

typo ... and fixed.

RalphSteinhagen:
> > assert(settings.get(TimingCtx(1, 1, 1, 2)).second == 102); // no setting for 'gid=2' exists yet -> fall back to 'gid=-1'
> 
> Such a silent fallback is rather dangerous and could result in setting wrong values to devices. Imo it would make more sense to dont do anything, for a device, and possibly signal a software interlock if a needed setting is missing. Though maybe I overlook something. How can I actually figure out that there is no setting provided for a specific gid ?

This fallback is on purpose and not silent an can be tested via:
```cpp
assert(settings.get(TimingCtx(1, 1, 1, 2)).first != TimingCtx(1, 1, 1, 2)); 
assert(settings.get(TimingCtx(1, 1, 1, 2)).first == TimingCtx(1, 1, 1)); 
```
The idea was that if there is, for example, no beam-process specific setting, that one gets the more general setting, i.e. with the ordering gid -> beam-process -> sequence -> chain -> non-multiplexed default. 

If such a behaviour is not needed that one could use the `BasicSetting<std::map<TimingCtx, T>>`.


> Sorry, I suppose so far all my comments are not directly related to this PR. If I should do them somewhere else, please let me know !

No worries, that's the purpose of this draft PR.

RalphSteinhagen:
@alexxcons thanks for your valuable feedback... this is proof that peer review works and is very useful!

> I think the stage/commit API is good, and I suppose would be sufiicient to cover FAIR transactional settings. Though I dont have the edge-cases for transactional settings in mind. It would be good if as well Dominic Day would review this PR (the API), since he is taking care for the transactional settings in Fesa.
> 

Contacted Dominic.... let's see what he and other think of the API.

> Some more questions:
> 
> * What happens if I stage twice for the same token ? I would expect an overwrite of the first staged value.

For the time being this is bit of UB :grimacing: : there would be two staged values that would be both commited (albeit ordered) if the token is submitted. Btw. the number/names of staged tokens can be queried via `getPendingTransactions()`.

> * What happens with data which is staged, but never commit ? I would expect that it just cycle out if the buffer rolls over.

There is a user-configurable time-out after which staged values are automatically dropped (trigged via `retireExpired()`) if there is no transaction token submitted. There is also the possibility using `bool retireStaged(const TransactionToken &transactionToken = NullToken<TransactionToken>)` to explicitly remove staged values.

The function `retireExpired()` removes also old history settings. There is also an 
```cpp
template<bool exactMatch = false>
bool retire(const TimingCtx &ctx) {..}
```
to explicitly retire settings for a given beam-process or chain/sequence and their dependent sub-beam-processes/sequences. The template argument controls whether the match is exact  e.g. `TimingCtx(1,1)` retiring only the Setting for 'cid=1 && sid=1' but leaving the depending 'cid=1 && sid=1 && pid= ...' intact, or `TimingCtx(1,1)` retiring all settings where the cid and sid matches.

dominicday:
With respect to multiple stagings, I would expect values to be overwritten and the timestamp of the final commit to be used. Is a commit without staging or a token the same as an immediate non-transactional setting elsewhere in opencmw?

Also, is it possible to request a timestamp for a commit? (eg so multiple devices have an identical timestamp matching a white rabbit event). Though this might also allow commits scheduled in the future.

For a comparison, with FESA there is one open transaction per value/selector-index pair. So 1 value can be prepared in advance for each of a series of beam processes. With only 1 slot, staged data that is not committed blocks the transaction buffer so the external transaction coordinator must keep track of which devices have accepted a transaction, ensuring each ID is retired before another is opened for the same BP.
The setting seems to be storing enough to be able to perform what the FAIR concept needs at the moment.

RalphSteinhagen:
Hi @dominicday thanks for your feedback

> With respect to multiple stagings, I would expect values to be overwritten and the timestamp of the final commit to be used. 
That's the effective behaviour but until the commit (with token) the stash would show both settings. Will address this somehow ... probably through filtering.

> Is a commit without staging or a token the same as an immediate non-transactional setting elsewhere in opencmw?
Yes, if there is not token, then the setting is committed immediately. N.B. the commit is intrinsically thread-safe and multiple threads can commit at the same time w/o locking. The one who effectively committed last will win though.

> Also, is it possible to request a timestamp for a commit? (eg so multiple devices have an identical timestamp matching a white rabbit event). Though this might also allow commits scheduled in the future.
Yes, the default return has the time-stamp of when the setting is committed as the second return argument of the structured binding, ie.
```cpp
auto [ok3, timeStamp3] = a.stage(53, "transactionToken#1");
```
Please note that this `timeStamp3` is w.r.t. the system time -- which using NTP/PTP -- should be only ~ 1 ms apart.
We (@wirew0rm  and I) have been discussing time-keeping/synchronisation a bit. There are failure cases both ways (ie. if using system or WR as a timing base). For the time-being, the line-of-though was/is to use the system time and to warn if this drifts by more then n-ms from the WR reference. The problem with the other way round is that the WR timing isn't always available w.r.t. 24h/7 365 days a year and also not available for all FECs. Thus NTP/PTP seems to be a safer fall back.
.. to be discussed what the best strategy is. 

> For a comparison, with FESA there is one open transaction per value/selector-index pair. So 1 value can be prepared in advance for each of a series of beam processes. With only 1 slot, staged data that is not committed blocks the transaction buffer so the external transaction coordinator must keep track of which devices have accepted a transaction, ensuring each ID is retired before another is opened for the same BP. The setting seems to be storing enough to be able to perform what the FAIR concept needs at the moment.

Yes, I was aware of the 'one open transaction per value/selector-index pair' at CERN which could be a limiting factor for FAIR where we have multiple trims/domains changing the same device.

Am glad that you think that this solution is compatible (/superset) of what is planned. I guess this will be also discussed as part of one of the upcoming CCT meetings.

What remains open is the specific API (naming, signature, etc.). Will wait a bit more for the other comments.

Thanks again for your feedback!

dominicday:
About the API:
Commit/stage looks good, structured binding is a nice way to return a pair.

For a newcomer, there are lots of get()s that behave differently (returns value or pair, finds approximate or exact match) which isn't obvious from the name. Probably OK for regular users though.

alexxcons:
> > > assert(settings.get(TimingCtx(1, 1, 1, 2)).second == 102); // no setting for 'gid=2' exists yet -> fall back to 'gid=-1'
> > 
> > 
> > Such a silent fallback is rather dangerous and could result in setting wrong values to devices. Imo it would make more sense to dont do anything, for a device, and possibly signal a software interlock if a needed setting is missing. Though maybe I overlook something. How can I actually figure out that there is no setting provided for a specific gid ?
> 
> This fallback is on purpose and not silent an can be tested via:
> 
> ```c++
> assert(settings.get(TimingCtx(1, 1, 1, 2)).first != TimingCtx(1, 1, 1, 2)); 
> assert(settings.get(TimingCtx(1, 1, 1, 2)).first == TimingCtx(1, 1, 1)); 
> ```
> 
> The idea was that if there is, for example, no beam-process specific setting, that one gets the more general setting, i.e. with the ordering gid -> beam-process -> sequence -> chain -> non-multiplexed default.
> 
> If such a behaviour is not needed that one could use the `BasicSetting<std::map<TimingCtx, T>>`.

Imo it is good to have the posibillity to get the setting of more general context, though if I request a very specific context, than I would expect to get the value for that very specific context. Getting something else by default is confusing, and probably would be a source of errors. ( I assume here, that the usual developer is lazy and wont double-check `setting.get(...).first` .. at least not until he/she runs into the problem :smile: )

I think I would do it like this:
- overload the(all) myField.get* methods with an enum 'use_fallback_if_not_found'. (Enum, because it is more descriptive than a true/false)
- If the default 'get' is used without enum, throw an exception, or return some kind of error (not sure how you used to handle failure)

Like that the feature would be opt-in, and dont catch developers by surprise.

RalphSteinhagen:
> I think I would do it like this:
> 
> * overload the(all) myField.get* methods with an enum 'use_fallback_if_not_found'. (enum over bool, because it is more descriptive than a true/false)
> * If the default 'get' is used without enum, throw an exception, or return some kind of error (not sure how you used to handle failure)

OK, w.r.t. to ENUM (or maybe just a boolean).

The greater API issue is how to communicate a missing setting. Throwing an exception for a relatively frequent use case isn't IMO a good design. Always using a `std::optional` has an indirection by default even for the case where there is always a return value returned (ie. matching or fall-back) and the returned value is also a `std::shared_ptr`. :thinking:

RalphSteinhagen:
Thanks again for the various feedbacks. I pushed some additional changes.

* replaced std::pair by more specific struct to hold return values, notably
```cpp
struct TransactionResult {
    using TimeStamp = std::chrono::system_clock::time_point;
    const bool                isCommitted;
    const TimeStamp           timeStamp;

    explicit(false) constexpr operator bool const &() const noexcept { return isCommitted; }
    explicit(false) constexpr operator TimeStamp const &() const noexcept { return timeStamp; }
    auto                      operator<=>(const TransactionResult &) const noexcept = default;
};

template<typename T>
struct CtxResult {
    const TimingCtx           timingCtx;
    const settings::node<T>  &settingValue;

    explicit(false) constexpr operator TimingCtx const &() const noexcept { return timingCtx; }
    explicit(false) constexpr operator T const &() const noexcept { return settingValue; }
    auto                      operator<=>(const CtxResult &) const noexcept = default;
};
```
* added explicit exception for requested timeStamp and indices that are out of range
* updated TimingCtx hash function

Also squashed the various commits to one reviewable one.

alexxcons:
> The greater API issue is how to communicate a missing setting. Throwing an exception for a relatively frequent use case isn't IMO a good design.

Practically I suppose that use-case here is, that LSA failed to supply settings before playing the pattern. In that case I suppose it is fine to get an execption (thoughl I might miss other use-cases)

alexxcons:
> > Imo it is good to have the posibillity to get the setting of more general context, though if I request a very specific context, than I would expect to get the value for that very specific context. Getting something else by default is confusing, and probably would be a source of errors. ( I assume here, that the usual developer is lazy and wont double-check `setting.get(...).first` .. at least not until he/she runs into the problem smile )
> 
> I would like to think about this less as a fallback as that if there is no specific setting, the more general setting _is_ the correct setting for that specific context. IMO if the user has to check whether he received a BP specific setting or a chain specific one points to the setting being not correctly modeled, as the client should just ask for a valid setting for his context, how this is resolved should in most cases be implementation detail. (Except of course for updating the setting, there it might be necessary to know where to overwrite.) Or maybe we need a less abstract example for where this would be a problem.

Yes, I suppose it would be a good idea to discuss a concrete example where using a fallback to a SID/GID makes sense if e.g. the requested PID is not available. Lets take a try:

Use case 1:  LSA failed to supply settings before playing the pattern. Our opencmw software is triggered on some WR-Event. the  GID/SID/PID of the event is used to search for some setting which is to apply to a device. 

- 1a) Since LSA did not set anything, we as well dont have a fallback-setting for the GID --> we go to some error state
- 1b) We still have a GID (or PID) setting from the last pattern which was played and use it, but it was wrong, beam is lost. (Is that a realistic use-case? That could as well happen today) Our device does not report any error.
- 1c) A device-expert, operator or user played with the FesaExplorer some while ago. Because of that, we have a fallback GID setting, which is now used. Our device makes use of that fallback (which sadly is outdated). The beam is lost because of that fallback, though our device did not report any error.

Use case 2: A client is requesting an acquisition value with an PID  for which we dont have data yet, because the PID was not played yet.

- 2a)  I suppose it does not make much sense to put fallback values for acquisition data. Maybe one could put "undefined" or "NaN" .. though if we want to rely on that, how to make sure that the buffer always has such a fallback available ?

Use case 3: A client is requesting a setting value with an PID for which we dont have data yet, because LSA did not supply settings yet.

- 3a) A device-expert, operator or user played with the FesaExplorer some while ago. Because of that, we have a fallback GID, which is returned now. The client might or might not check if the obtained value is a fallback.
- 3b) Nobody provided a fallback, so theclient receives an error.

I think 1c) and 3a) are the use-cases which worry me regarding automatic/silent fallback usage. Though what does not come to my mind is a use-case where a fallback actually would be helpfull/desired :thinking:  .. could you maybe provide something ?

RalphSteinhagen:
@alexxcons your concerns are noted. This setting implementation models different use-cases to those you described. 

> Practically I suppose the use-case here is, that LSA failed to supply settings before playing the pattern. In that case I suppose it is fine to get an execption (thoughl I might miss other use-cases)

No, that's not the primary use case and `opencmw::CtxSetting<T, ...>` is fundamentally different to the FESA modelling. If a FESA-like handling of settings (w/ all its pros and cons) and support for transactions is needed, one may opt to use:
```cpp
opencmw::BasicSetting<std::map<TimingCtx, T>>
```
which forces an exact matching. Especially, if you need multiplexing that is limited to a beam-processes-only scheme.

> Use case 2: A client is requesting an acquisition value with an PID for which we dont have data yet, because the PID was not played yet. [..]
> Use case 3: A client is requesting a setting value with an PID for which we dont have data yet, because LSA did not supply settings yet.

Both of these use-cases are not relevant because 'measurement data' isn't 'settings'.

---
What we modeled here is that for many devices the strict per-beam-process (BP) settings modeling is unnecessarily detailed, because their setting values are identical for a given sequence (S) of BPs, and/or chain of sequences (BPC), could be also non-multiplexed (NMP), and typically only require specialised settings only for a limited number of chains or sequences -- if at all..

The [TimingCtx](https://github.com/fair-acc/opencmw-cpp/blob/transactions/src/core/include/TimingCtx.hpp) models the WR timing context to consist of a tuple containing `<cid, sid, pid, bcts>` as further detailed here [here](https://git.gsi.de/acc/specs/opencmw#timing-multiplexing-modelling) and [here](https://www-acc.gsi.de/wiki/Timing/TimingSystemEvent). Also, as per LSA modeling, the tuple fields and contexts have a hierarchical dependence from '(NMP->)BPC->S->BP-> ...`. N.B. LSA doesn't model the transition between 'NMP' and multiplexed settings (yet).

---
The use-case that the `opencmw::CtxSetting<T, ...>` models works the other way round than what you described above: 
most devices are fine with only one static state captured by the type `T`. The generalised settings usually only needs to be specialised as needed and only at the level where it's physically meaningful. Notably: LSA and the FEC would need to store only one unique copy per resident BPC and multiplexing context (N.B. FESA pre-allocate and stores a copy for each possible multiplexing context, which are in the order of hundreds to thousands, where most often only a handful are needed).

For example: by default, regardless of the type of time-multiplexing, the `CtxSetting<T>::get()` would return always the default (global) value for a given device. This global setting can be changed globally via `FAIR.SELECTOR.ALL` (or the equivalent: `ALL`, `TimingCtx(cid=-1, sid=-1, pid=-1, gid=-1)`, ...).

If the same device should have the same setting for all BPCs except one, the setting could be specialised just, for example, for 
`TimingCtx(cid=-1)`. The call `CtxSetting<T>::get()` would then return always the global setting, except when the `TimingCtx` matches `cid=-1`. 

Another more accelerator-specific example: most diagnostic devices have the same setting per BPC (or sequence), this being the 95% use-case. For simplicity, let's assume it's a large setting that controls a computational-heavy transformation that needs to be applied on the raw data prior to being exported to the user. At FAIR, these beam-processes usually model distinct physics processes, for example, injection (`pid=1`), acceleration (`pid=2`), extraction (`pid=3`).
If a user requests only data during extraction, the selector would look like `TimingCtx(cid=..., sid=..., pid=3)`. Since there isn't a specific setting for this selector, one would need to either transform the whole chain and then filter the BP segment that matches the 'extraction beam process' (numerically heavy), or to retrieve the matching setting one-up in the hierarchy (here: BPC-multiplexed) and transform/return only the data for the requested BP. 

In case there is a need to adapt/specialise the setting just of the extraction beam-process, then the user would supply (typically via an expert-UI) only one modified setting for that particular beam-process w/o forcing all other setting values to be beam-process multiplexed by default.

---
A brief comment regarding expiring of old settings: FESA does not handle this at all and playing a new LSA context  w/o supplying new settings would wreck havoc with the same Timing Ctx.

That's the reason why we opted for `opencmw::BasicSetting<T, ...` and `opencmw::CtxSetting<T, ...` for two additional options for retiring old settings (and transactions for that matter):
  1. LSA overwrites the previously used TimingCtx with the new settings (status quo, FESA)
  1. via a configurable time-out: each `get()` updates the 'last-used' time stamp. If a setting hasn't been used/updated for an extended time period, it's automatically cleared and attached resources freed (N.B. suggested time-out would be on the scale of shifts to days)
  1. explicitly retiring settings if a resident chain gets removed from the resident timing schedule via:
```cpp
template<bool exactMatch = false>
bool retire(const TimingCtx &ctx, const TimeStamp &now = std::chrono::system_clock::now()) { ... }
```

alexxcons:
@RalphSteinhagen Thank you for the details and the examples !

> What we modeled here is that for many devices the strict per-beam-process (BP) settings modeling is unnecessarily detailed, because their setting values are identical for a given sequence (S) of BPs, and/or chain of sequences (BPC), could be also non-multiplexed (NMP), and typically only require specialised settings only for a limited number of chains or sequences -- if at all..

Alright,  I assumed that the current paradigm of settings values per BP would be continued. I have no idea on how many of these values are actually identical for specific devices. If it is the case, that values only deviate for a samll number of BPs, than for sure it makes much sense to define values per sequence (S) or chain (PBC) as "default" and only specify the few deviating ones per (BP).

( No idea if LSA is able to support that paradigm .. though if not, I suppose you have plans to patch LSA accordingly)

The main question for me is: Does it happen often, that  settings are the same for most BPs, but only are different for a small number ?

If so, than it makes sense to use the concept you suggest. Though if the standard use-case is to have a setting  per BP/Sequence /Chain than it would make sense for me to keep the current paradigm, and copy the setting to e.g. all but one  BP's for the exotic use-case in order to keep the API more simple and predictable.

> (N.B. FESA pre-allocate and stores a copy for each possible multiplexing context, which are in the order of hundreds to thousands, where most often only a handful are needed).

1024 slots per BP multiplexed field iirc. Though unlike you, Fesa is allocating the memory statically at startup. I suppose using opencmw, you could as well just dynamically allocate memory for the ~5 BP's which are actually used by LSA.

> Another more accelerator-specific example: most diagnostic devices have the same setting per BPC (or sequence), this being the 95% use-case. For simplicity, let's assume it's a large setting that controls a computational-heavy transformation that needs to be applied on the raw data prior to being exported to the user. At FAIR, these beam-processes usually model distinct physics processes, for example, injection (`pid=1`), acceleration (`pid=2`), extraction (`pid=3`). If a user requests only data during extraction, the selector would look like `TimingCtx(cid=..., sid=..., pid=3)`. Since there isn't a specific setting for this selector, one would need to either transform the whole chain and then filter the BP segment that matches the 'extraction beam process' (numerically heavy), or to retrieve the matching setting one-up in the hierarchy (here: BPC-multiplexed) and transform/return only the data for the requested BP.
> 
> In case there is a need to adapt/specialise the setting just of the extraction beam-process, then the user would supply (typically via an expert-UI) only one modified setting for that particular beam-process w/o forcing all other setting values to be beam-process multiplexed by default.

It is not clear to me how the new paradigm would help in that use-case. The large settings in 95% use-case still would be per chain if I get it correctly. So if I a client requests data for `pid=3`, than we only would be able to minimize CPU effort, if we indeed would have a dedicated setting for `pid=3` available.  Since we dont know what the client will request in advance, we would need to providce a setting for each pid again to minimize CPU effort.

I think it boils down to the same question than above: Is it a common use-case that the setting is the same for most BP's and only differs for few of them ? Than it makes sense.

> A brief comment regarding expiring of old settings: FESA does not handle this at all and playing a new LSA context w/o supplying new settings would wreck havoc with the same Timing Ctx.

Yes, it is nice to see posibillities to retire settings !  I discussed that with the LSA team years ago, though in the past there was not much interest, so I did not further follow that path.

wirew0rm (approved):
Looks good and I think the discussions on this PR also helped a lot to understand the different use cases.

I updated the sample code in the documentation to the changed struct return values and verified it actually compiles (Maybe from a maintainance pov it would be better to have this code as actual samples or testcases so it is checked by the CI and cannot become stale. Ideally there would be a solution to embed code from acutal samples into the documentation, but this should be discussed seperately and not block this PR).

I also silenced the remaining compilation warnings.

Two minor things I noticed is that `explicit(false)`does not suppress the clang-tidy in clion for me and also the code examples/tests cause some new clang compliation errors -> we really should get the clang CI running again soon.

@Ralph Please have a look at my 2 commits before (squas-)merging. The changes are not that big and rather uncontroversial, but a second pair of eyes always helps.

PR #163, opened 2022-05-10
Client Implementation

wirew0rm:
Client Implementation for the opencmw mdp(get/set) and mds(subscribe) protocol to be extended later by cmwLight and REST client implementations.

The design is based on the implementation in `opencmw-java` and consists of a thread that continuously polls all client implementations and a Commmand Ring Buffer which is used to allow requests from different threads.

### DIFF
--- a/src/client/CMakeLists.txt
+++ b/src/client/CMakeLists.txt
@@ -1,8 +1,8 @@
 # setup header only library
-add_library(client INTERFACE include/MockServer.hpp)
+add_library(client INTERFACE include/MockServer.hpp include/Client.hpp include/ClientContext.hpp)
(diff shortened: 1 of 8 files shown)


RalphSteinhagen (approved):
OK for first iteration. :+1:

things to be worked upon on the following iteration:
* composition: client choice is somewhat hard-coded and should be replaced/compostable by new user-provided implementations
* improve class naming, their documentation, and provide some end-user usage-examples
* round-trip performance benchmarking (e.g. 'client (get) -> simple property ->  client (reply)')
* complementary API to construct
  * RxCpp sources
  * Disruptor RingBuffer publishers

PR #164, opened 2022-05-12
basic thread pool that can optionally grow/shrink between [min, max] number of threads

RalphSteinhagen:
<h2>Basic thread pool that uses a fixed-number or optionally grow/shrink between a [min, max] number of threads.</h2>
The growth policy is controlled by the TaskType template parameter:
<ol type="A">
  <li> <code>TaskType::IO_BOUND</code> if the task is IO bound, i.e. it is likely to block the thread for a long time, or
  <li> <code>TaskType::CPU_BOUND</code> if the task is CPU bound, i.e. it is primarily limited by the CPU and memory bandwidth.
 </ol>
 <br>
 For the IO_BOUND policy, unused threads are kept alive for a pre-defined amount of time to be reused and gracefully
 shut down to the minimum number of threads when unused.
 <br>
 For the CPU_BOUND policy, the threads are equally spread and pinned across the set CPU affinity.
 <br>
 The CPU affinity and OS scheduling policy and priorities are controlled by:
 <ul>
  <li> <code>setAffinityMask(std::vector&lt;bool&gt; threadAffinityMask);</code> </li>
  <li> <code>setThreadSchedulingPolicy(const thread::Policy schedulingPolicy, const int schedulingPriority)</code> </li>
 </ul>

additionally includes:
 * AtomicMutex compatible with std::mutex and std::scoped_lock

### DIFF
--- a/src/core/include/ThreadPool.hpp
+++ b/src/core/include/ThreadPool.hpp
@@ -1,0 +1,381 @@
+#ifndef OPENCMW_CPP_THREADPOOL_HPP
+#define OPENCMW_CPP_THREADPOOL_HPP
+
+#include <atomic>
+#include <cassert>
+#include <chrono>
+#include <condition_variable>
+#include <list>
+#include <mutex>
+#include <span>
+#include <string>
+#include <thread>
+
+#include <fmt/format.h>
+
+#include <SpinWait.hpp>
+#include <ThreadAffinity.hpp>
+
+namespace opencmw {
+
+namespace thread_pool::detail {
+struct Task {
+    uint64_t              id;
+    std::function<void()> func;
+    Task                 *next = nullptr;
+    Task                 *self = this;
+    bool                  operator==(const Task &other) const { return self == other.self; }
+};
+
+class TaskQueue {
+    mutable AtomicMutex<> _lock;
+    Task                 *_head = nullptr;
+    Task                 *_tail = nullptr;
+    uint32_t              _size = 0;
+
+public:
+    TaskQueue()                       = default;
+    TaskQueue(const TaskQueue &queue) = delete;
+    TaskQueue &operator=(const TaskQueue &queue) = delete;
+    ~TaskQueue() { clear(); }
+
+    template<bool lock = true>
+    uint32_t clear() {
+        if constexpr (lock) {
+            _lock.lock();
+        }
+        const uint32_t res = _size;
+        for (Task *job = pop<false>(); job != nullptr; job = pop<false>()) {
+            delete job;
+        }
+        assert(_size == 0 && "TaskQueue::clear() failed");
+        if constexpr (lock) {
+            _lock.unlock();
+        }
+        return res;
+    }
+
+    uint32_t size() const {
+        std::scoped_lock lock(_lock);
+        return _size;
+    }
+
+    template<bool lock = true>
+    void push(Task *job) {
+        if constexpr (lock) {
+            _lock.lock();
+        }
+        job->next = nullptr;
+        if (_head == nullptr) {
+            _head = job;
+        }
+        if (_tail == nullptr) {
+            _tail = job;
+        } else {
+            _tail->next = job;
+            _tail       = job;
+        }
+        _size++;
+        if constexpr (lock) {
+            _lock.unlock();
+        }
+    };
+
+    template<bool lock = true>
+    Task *pop() {
+        if constexpr (lock) {
+            _lock.lock();
+        }
+        Task *head = _head;
+        if (head != nullptr) {
+            _head = head->next;
+            _size--;
+            if (head == _tail) {
+                _tail = nullptr;
+            }
+        }
+        if constexpr (lock) {
+            _lock.unlock();
+        }
+        return head;
+    };
+};
+
+} // namespace thread_pool::detail
+
+enum TaskType {
+    IO_BOUND  = 0,
+    CPU_BOUND = 1
+};
+
+template<typename T>
+concept ThreadPool = requires(T t, std::function<void()> &&func) {
+    { t.execute(std::move(func)) } -> std::same_as<void>;
+};
+
+/**
+ * <h2>Basic thread pool that uses a fixed-number or optionally grow/shrink between a [min, max] number of threads.</h2>
+ * The growth policy is controlled by the TaskType template parameter:
+ * <ol type="A">
+ *   <li> <code>TaskType::IO_BOUND</code> if the task is IO bound, i.e. it is likely to block the thread for a long time, or
+ *   <li> <code>TaskType::CPU_BOUND</code> if the task is CPU bound, i.e. it is primarily limited by the CPU and memory bandwidth.
+ * </ol>
+ * <br>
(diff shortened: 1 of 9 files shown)


src/core/include/ThreadPool.hpp
```diff
@@ -0,0 +1,341 @@
...
+            delete job;
+        }
+        assert(_size == 0 && "TaskQueue::clear() failed");
+        return res;
+    }
+
+    uint32_t size() const {
+        std::scoped_lock lock(_lock);
+        return _size;
+    }
```
mattkretz, line 54: What purpose does this function have? A caller gets some number returned. But at that point it has no useful meaning anymore. The actual size can be larger/smaller already. Without acquiring a lock from outside this function can just as well return `rand()`. :wink: And if you remove `size()` you can also remove `_size`.

src/core/include/ThreadPool.hpp
```diff
@@ -0,0 +1,341 @@
...
+    TaskQueue &operator=(const TaskQueue &queue) = delete;
+    ~TaskQueue() { clear(); }
+
+    uint32_t clear() {
+        const uint32_t res = _size;
+        for (Task *job = pop(); job != nullptr; job = pop()) {
+            delete job;
+        }
+        assert(_size == 0 && "TaskQueue::clear() failed");
+        return res;
```
mattkretz, line 48: Since this function doesn't hold the `_lock`, `res` is not the number of tasks that was cleared. (It can be, by chance, though.)
It seems like the function would be simpler and more efficient if you simply take the `_lock` and drain the linked list directly. If you decide to drop `_size`, just count `res`.

src/core/include/ThreadPool.hpp
```diff
@@ -0,0 +1,341 @@
...
+#include <ThreadAffinity.hpp>
+
+namespace opencmw {
+
+namespace thread_pool::detail {
+struct Task {
+    uint64_t              id;
+    std::function<void()> func;
+    Task                 *next = nullptr;
+    Task                 *self = this;
```
mattkretz, line 26: Why do you need `self`? I would expect you can always either use `this` or the `&` operator.

src/core/include/ThreadPool.hpp
```diff
@@ -0,0 +1,341 @@
...
+ * <ol type="a">
+ *   <li> the minimum/maximum thresholds, and</li>
+ *   <li> whether a task is scheduled via:</li>
+ *   <ol type="1">
+ *     <li> <code>enqueue(...)</code> i.e. re-using and queuing on an existing thread, or</li>
+ *     <li> <code>execute(...)</code> which may spawn a new thread if no free one is available.</li>
+ *   </ol>
+ * </ol>
+ * Unused threads are kept alive for a pre-defined amount of time to be reused and gracefully shut down to the
+ * minimum number of threads when unused.
```
mattkretz, line 107: Why? I.e. `enqueue` behaves differently depending on how many `execute` came before and depending on timing. Thus, `enqueue` can oversubscribe the processor or leave it mostly idle, unless the constructor is called with `#cores` for minimum and maximum threads.
It seems you want to be able to schedule *work* (which doesn't idle much) and *I/O* (which sleeps most of the time but shouldn't block the rest of the program).

typical requirements for *work*:
* use N0 cores, where `N0 <= #cores`
* queue the work until execution resources are available
* execute the work close to the data (NUMA awareness)

typical requirements for *I/O*:
* basically unbounded thread pool
* minimize latency until task is executed (avoid thread creation overhead)
* have N1 cores standing by exclusively for I/O

(`N0 + N1 + X == #cores`, where `X` often is 1 for the "main thread")

If that's what you want, I think you want a different interface.

mattkretz (commented):
I can continue review of `BasicThreadPool` if this is really the design you wanted.

src/core/include/ThreadPool.hpp
mattkretz: another requirement for *I/O*:
* execute in the NUMA domain of the device you're communicating with (e.g. PCI controller for Ethernet, IB, or GPU)

src/core/include/ThreadPool.hpp
RalphSteinhagen: The main purpose of `size()` is for the busy-loop to continue immediately and not to wait for the `_condition.wait(...)` being woken up.

The assumption is that `_size` is only modified/read with the lock `_lock` being held. 
For example, when `_size` is counted down in `clear()` this `size()` only returns when the `clear()` call has finished it's operation. In theory, this variable could perhaps be equally `std::atomic<int>`.

src/core/include/ThreadPool.hpp
RalphSteinhagen: The `self` is primarily used as a unique ID and to simplify the equals operator below that distinguishes task based on that. In one case a `Task` got copied and showed up being different even though the `id`, `func`, etc. (albeit not unique by themselves) were the same.

src/core/include/ThreadPool.hpp
RalphSteinhagen: Yes, lock is missing. :+1:

src/core/include/ThreadPool.hpp
RalphSteinhagen: @mattkretz basically, yes, this pool basically handles both of the use-cases that you outlined above and aims at minimising new and reusing existing threads that are a) named, b) bound to a given core, and c) have a given scheduling priority. The two cases are distinguished via:
  * `enqueue` is primarily for NUMA type loads (best-effort) and 
  * `execute` for potentially I/O blocking functions and spawns new threads if no existing free thread is available.

The number of [min,max] threads is configurable. The pool itself is configurable to spawn threads between a minimum number of long-lived threads and a maximum number of short-lived threads that may terminate if they haven't been used for a while.

Thought that one `ThreadPool` to rule both cases is possible ... or isn't?

RalphSteinhagen: @mattkretz we (@wirew0rm and I) had another brief discussion on this and what describes our signal-processing use-case best that we rapidly rotate through ideally non-overlapping 3 (+1) phases:

a) data-collection: IO-bound, firm real-time latencies (< 1ms, HW interrupt service routines, network IO, ...)
b) data-processing: NUMA-bound (e.g. spectral processing, linear algebra, AI, ...)
c) data-forwarding: IO-bound, firm real-time latencies (similar to a) but reverse)
d) basically sleep: handle user-IO, low-priority maintenance tasks (logging, gathering/writing metrics, etc.)

For most of our use-cases, the overall load is rather burst-like (100% for <10 ms) than continuous 100% CPU core usage. There are some cases where the load of 'b)' can be very low and the figure-of-merit is basically to minimise the group-delay between a) -> b).

RalphSteinhagen:
@mattkretz thanks again for your in-person review and feedback, update/split the implementation into `IO_BOUND` and `CPU_BOUND` thread pool policies.

adjusted the description above.

wirew0rm (approved):
Looks good :+1: API, test coverage and examples are very comprehensive.

PR #168, opened 2022-05-17
upgrade CI: ubuntu-2204(jammy), gcc: 11->12, clang: 13->14

wirew0rm:
Update the compiler versions on the CI to fix the clang build.


wirew0rm:
Reproducibly fails breaks `Rx timeout test`, but only in the CI for debug builds, locally the test runs through. Since this breaks coverage reporting we should fix this issue before merging this.

RalphSteinhagen (approved):
LGTM! :+1:

PR #169, opened 2022-05-17
fixed operator<<(..., Ringbuffer) overloading

RalphSteinhagen:
N.B. apparently gcc 12.1 defined a more strict (??) policy


wirew0rm (approved):
Very good, I also checked the build locally against my gcc 12.1 which now builds flawlessly! :+1:

PR #173, opened 2022-05-24
split and extended BasicThreadPool implementation [5]

RalphSteinhagen:
* pool implements (NTTP-controllable) [IO,CPU]-bound pools
* support std::future callbacks (~classic std::asynch interface)
* improved thread affinity/scheduling handling
* tasks can be (optionally) named
* tasks can be (optionally) priority-queued
* tasks can be (optionally) bound to a specific CPU that is compatible with the set affinity mask

Some basic usage examples (see BasicThreadPool_example.cpp for details):
```cpp
using namespace std::chrono;
using opencmw::thread::getThreadName;

opencmw::BasicThreadPool<opencmw::CPU_BOUND> poolWork("CustomCpuPool", 1, 1); // pool for CPU-bound tasks with exactly 1 thread
opencmw::BasicThreadPool<opencmw::IO_BOUND>  poolIO("CustomIOPool", 1, 1000); // pool for IO-bound (potentially blocking) tasks with at least 1 and a max of 1000 threads
poolIO.keepAliveDuration() = seconds(10);                                     // keeps idling threads alive for 10 seconds
poolIO.waitUntilInitialised();                                                // wait until the pool is initialised (optional)
assert(poolIO.isInitialised());                                               // check if the pool is initialised

// enqueue and add task to list -- w/o return type
poolWork.execute([] { fmt::print("Hello World from thread '{}'!\n", getThreadName()); });
poolWork.execute([](const auto &...args) { fmt::print("Hello World from thread '{}'!\n", args...); }, getThreadName());

constexpr auto           func1  = [](const auto &...args) { return fmt::format("thread '{1}' scheduled task '{0}'!\n", getThreadName(), args...); };
std::future<std::string> result = poolIO.execute<"customTaskName">(func1, getThreadName());
// do something else ... get result, wait if necessary
std::cout << result.get() << std::endl;
poolIO.setAffinityMask({ true, true, true, false });
poolIO.execute<"task name", 20U, 2>([]() { fmt::print("Hello World from custom thread '{}'!\n", getThreadName()); }); // execute a task with a name, a priority and single-core affinity
try {
    poolIO.execute<"customName", 20U, 3>([]() { /* this potentially long-running task is trackable via it's 'customName' thread name */});
} catch (const std::invalid_argument &e) {
    fmt::print("caught exception: {}\n", e.what());
}
```

Example output:
```
thread 'BasicThreadPool' scheduled task 'customTaskName'!

Hello World from thread 'CustomCpuPool#1'!
Hello World from custom thread 'task name'!
Hello World from thread 'BasicThreadPool'!
caught exception: requested cpuID 3 incompatible with set affinity mask(4): [true, true, true, false]
```

### DIFF
--- a/src/core/include/ThreadPool.hpp
+++ b/src/core/include/ThreadPool.hpp
@@ -5,26 +5,116 @@
 #include <cassert>
 #include <chrono>
 #include <condition_variable>
+#include <deque>
+#include <functional>
+#include <future>
+#include <iostream>
 #include <list>
 #include <mutex>
 #include <span>
 #include <string>
 #include <thread>
+#include <type_traits>
+#include <utility>
 
 #include <fmt/format.h>
+#include <fmt/ranges.h>
 
+#include <opencmw.hpp>
 #include <SpinWait.hpp>
 #include <ThreadAffinity.hpp>
 
 namespace opencmw {
 
+class TaskQueue;
+
 namespace thread_pool::detail {
+
+/**
+ * @brief a move-only implementation of std::function
+ * originally from: https://stackoverflow.com/questions/25330716/move-only-version-of-stdfunction#answer-52358928
+ * to be replaced once C++23 is out/available.
+ * For details see:
+ * https://en.cppreference.com/w/cpp/utility/functional/move_only_function/move_only_function
+ */
+template<typename T>
+class unique_function : public std::function<T> {
+    template<typename Fn>
+    struct wrapper;
+
+    template<std::copy_constructible Fn>
+    struct wrapper<Fn> {
+        Fn fn;
+
+        template<typename... Args>
+        auto operator()(Args &&...args) { return fn(FWD(args)...); }
+    };
+
+    template<std::move_constructible Fn>
+    requires(!std::is_copy_constructible_v<Fn>) struct wrapper<Fn> {
+        Fn fn;
+
+        explicit wrapper(Fn &&func) noexcept
+            : fn(std::move(func)) {}
+
+        wrapper(wrapper &&) noexcept = default;
+        wrapper &operator=(wrapper &&) noexcept = default;
+
+        // these two functions are instantiated by std::function and are never called
+        wrapper(const wrapper &rhs)
+            : fn(const_cast<Fn &&>(rhs.fn)) { throw 0; } // hack to initialize fn for non-DefaultConstructible types
+        wrapper &operator=(wrapper &) { throw 0; }
+
+        template<typename... Args>
+        auto operator()(Args &&...args) { return fn(FWD(args)...); }
+    };
+
+    using base = std::function<T>;
+
+public:
+    using base::operator();
+
+    unique_function() noexcept = default;
+    explicit unique_function(std::nullptr_t) noexcept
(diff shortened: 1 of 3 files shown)


src/core/include/ThreadPool.hpp
```diff
@@ -5,26 +5,116 @@
...
+
+        explicit wrapper(Fn &&func) noexcept
+            : fn(std::move(func)) {}
+
+        wrapper(wrapper &&) noexcept = default;
+        wrapper &operator=(wrapper &&) noexcept = default;
+
+        // these two functions are instantiated by std::function and are never called
+        wrapper(const wrapper &rhs)
+            : fn(const_cast<Fn &&>(rhs.fn)) { throw 0; } // hack to initialize fn for non-DefaultConstructible types
```
mattkretz, line 65: Maybe only
```c++
wrapper(const wrapper &rhs);
```
without definition? I.e. make it fail when linking rather than at runtime.

src/core/include/ThreadPool.hpp
```diff
@@ -5,26 +5,116 @@
...
+        wrapper &operator=(wrapper &) { throw 0; }
```
mattkretz, line 66: same here, of course

src/core/include/ThreadPool.hpp
RalphSteinhagen: @mattkretz unfortunately, the version with `wrapper(const wrapper &rhs);` does not compile. 

Part of the 'hackish' implementation is that this class derives from `std::function(...)` and needs to shadow the superordinate copy-asignment/constructor implementation.

mattkretz: Too bad. Maybe you want to consider a special-purpose `function` type like in https://godbolt.org/z/4GzGoh68z?

mattkretz (approved):
LGTM

src/core/include/ThreadPool.hpp
```diff
@@ -5,26 +5,85 @@
...
-    bool                  operator==(const Task &other) const { return self == other.self; }
+    uint64_t            id;
+    function            func;
+    std::string         name     = "";
+    int32_t             priority = 0;
+    int32_t             cpuID    = -1;
+    Task               *next     = nullptr;
+    Task               *self     = this;
+    bool                operator==(const Task &other) const noexcept { return self == other.self; }
+    auto                operator<=>(const Task &other) const noexcept { return priority <=> other.priority; }
```
mattkretz, line 78: This is strange. You can have `a == b` be false while `is_eq(a <=> b)` is true. `auto` in this case will be `std::strong_ordering`. But what you actually seem to want is a partial ordering (like with NaNs in floating-point).

mattkretz (changes requested):
I'd be afraid of the `Task` comparison operators biting at some point. And it'll probably be mind-boggling why results are inconsistent.

src/core/include/ThreadPool.hpp
RalphSteinhagen: @mattkretz removed `operator==` wasn't called even for tasks with matching priorities.

mattkretz (approved):

PR #207, opened 2022-07-28
Add header for cross access

johannaheuser:
The header file did not provide access control allowance so that responds can be shared with requesting code. Further information on Access-Control-Allow-Origin can be found [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin).


wirew0rm (approved):
Looks good.

In a later commit we might want to make this configurable.
If I understand everything correct, this should not be necessary any more once the wasm gui is also served by the majordomo rest server. On the other hand, clients might request data from different brokers.

PR #208, opened 2022-08-16
Add support for HTTP OPTIONS preflight request

wirew0rm:
To do long polling the opencmw rest protocol requires a special
X_OPENCMW_METHOD header which disables the simple CORS handling and
requires the server to respond to a pre-flight request [1].
This commit adds support for preflight requests allowing access with
the custom opencmw headers from any origin.
Builds up on #207, which allows simple CORS requests without the
headers.

This is not needed if the client html is served from the same origin
(protocol, host, port) which is why the problem is not happening with
the html based webinterface.

Additionally, the the `Acces-Control-Allow-Origin` header was only set
for get responses and not for subscription responses, which is also
fixed in this commit.

[1] https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request


johannaheuser (commented):
I tested the changes with the emscripten fetch API. For that case, everything works fine.

johannaheuser (approved):
See above.

PR #210, opened 2022-08-31
fix compilation for emscripten with certain specific hacks

PuerNoctis:
partially fixing #193 (only Emscripten/clang)

The current stable emsdk ships with an STL implementation that, in combination with using clang/LLVM, is incompatible with refl-cpp and mp-units. To fix compilation, some direct hacks have been introduced. They seem to work without breaking functionality with and without using Emscripten, but when something changes in emsdk, there might be new issues.

Another idea would be to try and see if emsdk has unstable releases of a newer STL and/or better C++20 support. Overall however, this PR shows what _would_ need to be done, to get opencmw compiled using Emscripten/Clang.

### DIFF
--- a/src/core/include/opencmw.hpp
+++ b/src/core/include/opencmw.hpp
@@ -12,13 +12,42 @@
 #include <fmt/format.h>
 #include <fmt/ostream.h>
 #include <fmt/ranges.h>
-#include <refl.hpp>
+
+// refl-cpp uses forward declarations for smart-pointers that are incompatible
+// with current emscripten STL headers. The "fix" is a little hack to temporarily
+// define __APPLE__ which causes refl-cpp to not forward-declare the classes.
+// Seems to work for now.
+#ifdef __EMSCRIPTEN__
+#   define __APPLE__
+#   include <refl.hpp>
+#   undef __APPLE__
+#else
+#   include <refl.hpp>
+#endif
+
 #pragma GCC diagnostic push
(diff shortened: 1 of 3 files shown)


wirew0rm:
Great work! I would like to involve some more people into the discussion and do some testing before deciding whether to merge this. Open question for me is if we want to remove `long long` for all platforms or only for emscripten.

> Another idea would be to try and see if emsdk has unstable releases of a newer STL and/or better C++20 support.

I found: https://github.com/emscripten-core/emscripten/issues/17680 -> they use different versions of clang vs their libc++ fork and for updating libc++ some patches have to be applied and conflicts resolved. Currently they are on libc++ 14, since there are only pre-release versions of llvm 15.

PuerNoctis:
The problem with `long long` at my end was, that `uint64_t` is `unsigned long` on systems with LP64, and `unsigned long long` on LLP64. The latter of which caused a conflicting overload with the actual `long long` variants. Alternatively we could _NOT_ use the typedefs from stdint.h.

Overall I would also agree to have a discussion and some further tests on this PR. The solution is indeed quite hacky - an updated version of libc++ would make me feel more comfortable. On the other hand we would have to check whether this also solves the issues with e.g. mp-units.

ivan-cukic:
This looks mostly Ok considering "hacks" in the title. :)

Ideally, if the Emscripten specifics start to grow, I'd prefer everything to be moved inside something header so that we could have 'platform_std.h', 'platform_emscripten.h' or such.

RalphSteinhagen:
resolved through PRs
  * #235 
  * #238

PR #212, opened 2022-09-08
exposed _activeSubscriptions in BasicWorker

RalphSteinhagen:
should be thread-safe and allow for custom notification based on existing/known subscriptions.


wirew0rm (approved):
:+1: Looks good.

I think it would be nice to have a sample that shows how to use this, but that ca be done in a separate PR

PR #215, opened 2022-09-08
Mustache MultiArray

wirew0rm:
Adds support for domain objects containing MultiArray to the MustacheSerialiser.

### DIFF
--- a/src/serialiser/include/MustacheSerialiser.hpp
+++ b/src/serialiser/include/MustacheSerialiser.hpp
@@ -36,9 +36,6 @@ template<typename T>
 class mustache_data;
 
 template<typename T>
-using mustache = kainjow::mustache_ns<std::string, mustache_data<T>>;
-
-template<typename T>
 constexpr bool is_instance_of_vector_v = units::is_derived_from_specialization_of<T, std::vector>;
 static_assert(!is_instance_of_vector_v<int>);
 static_assert(!is_instance_of_vector_v<MultiArray<double, 2>>);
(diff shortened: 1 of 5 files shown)


src/serialiser/include/MustacheSerialiser.hpp
```diff
@@ -36,7 +36,18 @@ template<typename T>
 class mustache_data;
 
 template<typename T>
-using mustache = kainjow::mustache_ns<std::string, mustache_data<T>>;
+constexpr bool is_instance_of_vector_v = units::is_derived_from_specialization_of<T, std::vector>;
```
RalphSteinhagen, line 39: There is already an existing identical definition `opencmw::is_vector` in [opencmw.hpp](https://github.com/fair-acc/opencmw-cpp/blob/185398eacc1028bebe4452bfa31fd5e09dca533f/src/core/include/opencmw.hpp#L116-L117) which is included above and also handles the `const` case. Maybe reuse that.

src/serialiser/include/MustacheSerialiser.hpp
```diff
@@ -36,7 +36,18 @@ template<typename T>
...
+static_assert(!is_instance_of_vector_v<int>);
+static_assert(!is_instance_of_vector_v<MultiArray<double, 2>>);
+static_assert(is_instance_of_vector_v<std::vector<double>>);
+
+template<typename T>
```
RalphSteinhagen, line 44: There is already a similar SFINAE definition in [MultiArray.hpp](https://github.com/fair-acc/opencmw-cpp/blob/185398eacc1028bebe4452bfa31fd5e09dca533f/src/serialiser/include/MultiArray.hpp#L297-L304). Probably to re-use that . I like the static_asserts compile-time checks. Maybe they could be moved to that header?

RalphSteinhagen (approved):
Thanks for the quick fix. :+1:

Good as is and could be merged, but I'd recommend to nevertheless eliminate the dead-code and re-use the type definition for `is_vector` in opencmw.hpp and `is_multi_array` in MultiArray.hpp. This keeps the option to adapt/optimise those definitions globally, if necessary.

src/serialiser/include/MustacheSerialiser.hpp
wirew0rm: changed the SFINAE definiton in MultiArray to this one and removed it here. Static asserts moved to MultiArray_tests.

src/serialiser/include/MustacheSerialiser.hpp
wirew0rm: done

PR #218, opened 2022-09-08
Client: keep query params

wirew0rm:
Make the client use the path and the query to make subscriptions as opposed to only the path. This is required to allow subscriptions with different parameters/filters to be notified.

Also cleans up some warnings and other small things.

### DIFF
--- a/src/client/include/Client.hpp
+++ b/src/client/include/Client.hpp
@@ -96,26 +96,26 @@ public:
 
     void get(const URI<STRICT> &uri, majordomo::MessageFrame &req_id) override {
         detail::Connection &con     = findConnection(uri);
-        auto                message = createRequestTemplate(majordomo::Command::Get, uri.path().value(), req_id);
+        auto                message = createRequestTemplate(majordomo::Command::Get, uri.relativeRefNoFragment().value(), req_id);
         message.send(con._socket).assertSuccess();
     }
 
     void set(const URI<STRICT> &uri, majordomo::MessageFrame &req_id, const std::span<const std::byte> &request) override {
         auto &con     = findConnection(uri);
-        auto  message = createRequestTemplate(majordomo::Command::Set, uri.path().value(), req_id);
+        auto  message = createRequestTemplate(majordomo::Command::Set, uri.relativeRefNoFragment().value(), req_id);
         message.setBody(std::string(reinterpret_cast<const char *>(request.data()), request.size()), majordomo::MessageFrame::dynamic_bytes_tag{});
         message.send(con._socket).assertSuccess();
     }
 
     void subscribe(const URI<STRICT> &uri, majordomo::MessageFrame &req_id) override {
         auto &con     = findConnection(uri);
-        auto  message = createRequestTemplate(majordomo::Command::Subscribe, uri.path().value(), req_id);
+        auto  message = createRequestTemplate(majordomo::Command::Subscribe, uri.relativeRefNoFragment().value(), req_id);
         message.send(con._socket).assertSuccess();
     }
 
     void unsubscribe(const URI<STRICT> &uri, majordomo::MessageFrame &req_id) override {
         auto &con     = findConnection(uri);
-        auto  message = createRequestTemplate(majordomo::Command::Unsubscribe, uri.path().value(), req_id);
+        auto  message = createRequestTemplate(majordomo::Command::Unsubscribe, uri.relativeRefNoFragment().value(), req_id);
         message.send(con._socket).assertSuccess();
     }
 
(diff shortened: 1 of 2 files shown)


src/core/include/URI.hpp
```diff
@@ -169,6 +182,7 @@ class URI {
     inline std::optional<uint16_t> port() const noexcept { parseAuthority(); return _port.empty() ? std::nullopt : std::optional(std::stoi(std::string{ _port.begin(), _port.end() }));}
     inline std::optional<std::string> path() const noexcept { return returnOpt(_path); }
     inline std::optional<std::string> queryParam() const noexcept { return returnOpt(_query); }
+    inline std::optional<std::string> localPart() const noexcept { return returnOpt(merge_adjacent_string_views({_path.data(), _path.size()+1}, _query)); } // path + query
```
RalphSteinhagen, line 185: N.B. this should work for well-formatted URIs.

I like the definition of this `localPart()` because there is a recurring need. I wonder whether we shouldn't also include the fragment `#...` part of the URI. The server may (/should) drop this for most cases.
However, since we use this also internally to our application the fragment can convey important app-specific information.

RalphSteinhagen (changes requested):
some minor changes to be discussed: 
* whether `URI::localPart()` should return also the fragment, ie. `<path>?<query_parm>#<fragment>` 
* naming: `URI::localPart()` to `URI::relativeRef()` as described in [rfc3986 - Section 4.2](https://www.rfc-editor.org/rfc/rfc3986#section-4.2)

Otherwise, LGTM! :+1:

src/client/include/Client.hpp
```diff
@@ -243,20 +248,21 @@ class SubscriptionClient : public MDClientBase {
...
-        auto            &con         = findConnection(uri);
-        std::string_view serviceName = uri.path().value();
+    void subscribe(const URI<STRICT> &uri, majordomo::MessageFrame & /*reqId*/) override {
+        auto            &con               = findConnection(uri);
+        std::string      serviceNameString = uri.localPart().value();
+        std::string_view serviceName       = serviceNameString;
         serviceName.remove_prefix(std::min(serviceName.find_first_not_of('/'), serviceName.size()));
         assert(!serviceName.empty());
-        opencmw::majordomo::zmq_invoke(zmq_setsockopt, con._socket, ZMQ_SUBSCRIBE, serviceName.data(), serviceName.size()).assertSuccess();
+        opencmw::majordomo::zmq_invoke(zmq_setsockopt, con._socket, ZMQ_SUBSCRIBE, serviceName.data(), serviceName.size() - 1).assertSuccess(); // string_view size includes termination byte
```
RalphSteinhagen, line 257: This produces an error because the string_view not actually being NULL-terminated `\0`. 

Using a copy-to-string workaround like:
```cpp
        std::string serviceName2(serviceName);
        opencmw::majordomo::zmq_invoke(zmq_setsockopt, con._socket, ZMQ_SUBSCRIBE, serviceName2.data(), serviceName2.size()).assertSuccess(); // string_view size includes termination byte
```
seems to fix this but should look into a more elegant version.

src/client/include/Client.hpp
```diff
@@ -243,20 +248,21 @@ class SubscriptionClient : public MDClientBase {
...
 
-    void unsubscribe(const URI<STRICT> &uri, majordomo::MessageFrame &reqId) override {
+    void unsubscribe(const URI<STRICT> &uri, majordomo::MessageFrame & /*reqId*/) override {
         auto            &con         = findConnection(uri);
-        std::string_view serviceName = uri.path().value();
+        std::string_view serviceName = uri.localPart().value();
         serviceName.remove_prefix(std::min(serviceName.find_first_not_of('/'), serviceName.size()));
         assert(!serviceName.empty());
-        opencmw::majordomo::zmq_invoke(zmq_setsockopt, con._socket, ZMQ_UNSUBSCRIBE, serviceName.data(), serviceName.size()).assertSuccess();
+        opencmw::majordomo::zmq_invoke(zmq_setsockopt, con._socket, ZMQ_UNSUBSCRIBE, serviceName.data(), serviceName.size() - 1).assertSuccess();
```
RalphSteinhagen, line 265: as above for the 'subscribe' version.

src/client/include/Client.hpp
wirew0rm: This was due to an error in my implementation of `localPart()` with empty query returning one character too much. Removed the -1 part.

src/core/include/URI.hpp
wirew0rm: Renamed to `relativeRefNoFragment()` and introduced an additional `relativeRef()`.

RalphSteinhagen (approved):
As discussed and thanks for the changes! LGTM!

PR #226, opened 2022-09-15
Upgrade libfmt

wirew0rm:
upgrades libfmt to 8.1.1

open questions:
- should we go directly to v9?
- formatter instead of static casts for std::errc?
- use of fmt::runtime in ThreadPool sample

see comments in the individual commits for more details


RalphSteinhagen (approved):
LGTM! Should clean up some of the dead-code commented above

PR #228, opened 2022-09-16
added FilterSubscription example

RalphSteinhagen:
to illustrate a more advanced notify scheme with clients providing a query-parameter-based filter

### DIFF
--- a/concepts/majordomo/FilterSubscription_example.cpp
+++ b/concepts/majordomo/FilterSubscription_example.cpp
@@ -1,0 +1,178 @@
+#include <Client.hpp>
+#include <majordomo/base64pp.hpp>
+#include <majordomo/Broker.hpp>
+#include <majordomo/RestBackend.hpp>
+#include <majordomo/Worker.hpp>
+
+#include <atomic>
+#include <fstream>
+#include <iomanip>
+#include <math.h>
+#include <ranges>
+#include <string_view>
+#include <thread>
+
+using namespace opencmw::majordomo;
+using namespace std::chrono_literals;
+
+struct FilterContext {
+    std::string             signalFilter;
+    opencmw::MIME::MimeType contentType = opencmw::MIME::BINARY;
+};
+ENABLE_REFLECTION_FOR(FilterContext, signalFilter, contentType)
+
+struct Reply {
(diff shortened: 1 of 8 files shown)


concepts/majordomo/FilterSubscription_example.cpp
```diff
@@ -0,0 +1,177 @@
...
+    std::atomic<int>               receivedA{ 0 };
+    std::atomic<int>               receivedAB{ 0 };
+    clientContextA.subscribe(URI("mds://127.0.0.1:12345/DeviceName/Acquisition?signalFilter=A"), [&receivedA](const opencmw::mdp::Message &update) {
+        fmt::print("Client('A') received message from service '{}' for endpoint '{}'\n", update.serviceName.str(), update.endpoint.str());
+        receivedA++;
+    });
+    clientContextA.subscribe(URI("mds://127.0.0.1:12345/DeviceName/Acquisition?signalFilter=A,B"), [&receivedAB](const opencmw::mdp::Message &update) {
+        fmt::print("Client('A,B') received message from service '{}' for endpoint '{}'\n", update.serviceName.str(), update.endpoint.str());
+        receivedAB++;
+    });
```
sfalamic, line 164: Here we subscribe but the clients actually do not receive data from the worker.

sfalamic (commented):
This looks good! Are we expecting that the clients actually receive data from the worker?

concepts/majordomo/FilterSubscription_example.cpp
RalphSteinhagen: That's why we make these PR reviews!

@sfalamic thanks for spotting this! Will look into this.

sfalamic: > Should be fixed now, was an error in the WIP OpenCMW client implementation that suppressed the leading '/' of the subscription.

Thank you for the fix. The OpenCMW clients now receive messages from the worker.

> N.B. this problem is specific to the 'Client' implementation. For the REST client this isn't an issue.

Exactly, for the REST client, it already worked as expected.

sfalamic (approved):
Looks good now!

PR #239, opened 2022-12-01
fixed JSON formatter

RalphSteinhagen:
... to not emit a leading map name for the outer scope.

was: '"<root>": { /*[..]*/ }'
now: '{ /*[..]*/ }'

produces JSON compatible with https://jsonlint.com/ et. al.


ivan-cukic (approved):
LGTM

Issue #231, opened 2022-10-20
[0pt] Optimize REST backend to work with several clients


ivan-cukic:
1. Keep subscriptions open for a specified number of seconds after the REST client goes away to remove reconnecting if a new client pops up that wants the same service (or if the same client sends a new request)
2. Allow having a single client that requests several services
3. Add 'persistent' URLs for caching clients

RalphSteinhagen:
## some additional information:
one of the OpenCMW goals is that recurring/similar REST user requests can be cached by transparent HTTP proxies (external to OpenCMW, e.g. [Squid](http://www.squid-cache.org/)) to optimise the actual required bandwidth/cpu usage on the service, to efficiently fan-out the same information to multiple clients (e.g. fixed-display applications/dashboards), and is also essential to be able to safely forward/route 'GET' requests to unsafe/public networks.

The issue/need for the modification is that most caching HTTP proxies by-default only cache 'GET' responses and primarily rely on the URI and MIME-type as the cache key. For further details see, for example, [RFC 9111](https://httpwg.org/specs/rfc9111.html#rfc.section.2.p.3):
> [..] However, many HTTP caches in common use today only cache GET responses and therefore only use the URI as the cache key.

and Mozilla's [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#cache_busting) on cache busting:
> [..] And for resources that do change, it is a common best practice to change the URL each time the content changes, so that the URL unit can be cached for a longer period.

### query parameter implementation detail
Thus, in complementary to using [SSE](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events), the 'RestBackend' would need to be modified to support long-polling using an additional query parameter (proposal: `LongPollingIdx`) with a  unique counter for each notification/subscription e.g.
`http://.../path?signals=A,B,C&LongPollingIdx=0`,
`http://.../path?signals=A,B,C&LongPollingIdx=1`,
`http://.../path?signals=A,B,C&LongPollingIdx=2`,
...
`http://.../path?signals=A,B,C&LongPollingIdx=7`,
`http://.../path?signals=A,B,C&LongPollingIdx=8`
[...]

PR #245, opened 2022-12-19
Reimplemented long pollin and subscription for the REST backend

ivan-cukic:


RalphSteinhagen (approved):
Thanks for upgrading the REST server. This addresses issue [#231](https://github.com/fair-acc/opencmw-cpp/issues/231)

The clang compile error indicated in https://github.com/fair-acc/opencmw-cpp/actions/runs/3759535466/jobs/6389216117#step:12:1277
could be fixed via using
```cpp
std::atomic_flag _lock = ATOMIC_FLAG_INIT;
```

Otherwise OK and could be merged once also the clang CI compiles. :+1:

PR #259, opened 2023-02-23
RestClient: 200 is not an HTTP error code

giucam:


wirew0rm (approved):
Thanks for catching and fixing this!

Failing tests for clang16 and emscripten are due to CI errors already fixed in #258 and unrelated.

PR #264, opened 2023-04-05
Build/link fixes for the client module

giucam:


wirew0rm:
closed in favor of marking the methods as inline.

PR #268, opened 2023-04-06
Allow Content-Type as a header for Cross-Origin Resource Sharing (CORS)

giucam:


wirew0rm (approved):
:+1: 

By default contentType is allowed for CORS requests but only supports a very basic set of values. Adding it as a supported header explicitly lifts these [restrictions](
https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header#additional_restrictions).

PR #271, opened 2023-04-07
Allow making REST requests to other domains on emscripten

giucam:


giucam:
This needs #268 too to be useful

wirew0rm (approved):
:+1:

PR #273, opened 2023-04-13
Allow a worker at /Path to handle /Path/... as well

giucam:


src/majordomo/test/majordomo_tests.cpp
```diff
@@ -581,6 +581,132 @@ TEST_CASE("One client/one worker roundtrip", "[broker][roundtrip]") {
...
+    ready.setServiceName("/DeviceA/dashboard", static_tag);
+    ready.setBody("An example worker serving different dashbards", static_tag);
+    ready.setRbacToken("rbacToken", static_tag);
+    worker.send(ready);
+
+    broker.processMessages();
+
+    {
+        auto request = MdpMessage::createClientMessage(Command::Get);
+        request.setServiceName("/DeviceA/dashboard", static_tag);
```
wirew0rm, line 606: might be that service name should be `DeviceA/dashboard` (without the leading slash)

src/majordomo/test/majordomo_tests.cpp
```diff
@@ -581,6 +581,132 @@ TEST_CASE("One client/one worker roundtrip", "[broker][roundtrip]") {
...
+
+    // verify that the broker shuts the worker down correctly
+    const auto disconnect = worker.tryReadOneSkipHB(3);
+    REQUIRE(disconnect.has_value());
+    REQUIRE(disconnect->isValid());
+    REQUIRE(disconnect->isWorkerMessage());
+    REQUIRE(disconnect->command() == Command::Disconnect);
+    REQUIRE(disconnect->serviceName() == "/DeviceA/dashboard");
+    REQUIRE(disconnect->clientRequestId().empty());
+    REQUIRE(disconnect->topic() == "//DeviceA/dashboard");
```
wirew0rm, line 704: The additional leading slash is appended to the service name by the broker.
Probably the worker implementations also do that for the topic, which would mean that the service name is without leading slash and the topic is `/<serviceName>[<optional suffix>]?[<optional query parameters>]`.

Also in the requests service name is with the suffix, but the reply only sets the base service name.

wirew0rm (commented):
I've added a regression test with the intended behavior.

There are still some inconsistencies with relation between service name and topic, mostly regarding the leading slash, which I'll have to fix either in the test if the current behavior is ok or in the code if this should be improved.

wirew0rm (approved):
Good for now, pending the unification of the service names and topics throughout the project which will be done in a separate PR

PR #274, opened 2023-04-17
Fixes for subscriptions with REST 

giucam:


src/client/include/RestClientEmscripten.hpp
```diff
@@ -156,7 +156,7 @@ struct SubscriptionPayload : FetchPayload {
             payload->onerror(fetch->status, std::string_view(fetch->data, detail::checkedStringViewSize(fetch->numBytes)), fetch->statusText);
             emscripten_fetch_close(fetch);
         };
-        emscripten_fetch(&attr, uri.relativeRef()->data());
+        emscripten_fetch(&attr, uri.str().data());
```
ivan-cukic, line 159: Intentional change?

wirew0rm: This seems to have been missed in https://github.com/fair-acc/opencmw-cpp/pull/271/files#diff-3cdd884b6f9b6dbe7707176330fe745e7b232f71f81fa5b09851a4aac653844eR268 and to me it seems logical that otherwise subscriptions would only work on the same server or is the rest of the url passed to fetch somewhere else?

giucam: Ineed i had missed it there because I hand't tried to make a subscription yet and I didn't realize it was a different codepath. The erst of the url is not pased somewhere else, as far as i can see, subscriptions to different servers are  indeed not working without this.

giucam:
I don't understand why there is a failing test on ubuntu 22.04 debug. It does not fail locally here nor on the other ones.

src/majordomo/include/majordomo/RestBackend.hpp
```diff
@@ -742,10 +742,10 @@ struct RestBackend<Mode, VirtualFS, Roles...>::RestWorker {
     bool respondWithLongPoll(const httplib::Request &request, httplib::Response &response, const std::string_view &_service) {
         // TODO: After the URIs are formalized, rethink service and topic
         auto                     split = std::ranges::find(_service, '/');
-        std::string_view         service(_service.begin(), split);
-        std::string_view         topic(split, _service.end());
+        std::string              service(_service.begin(), split);
+        std::string              topic = "/"s + service + std::string(split, _service.end());
```
wirew0rm, line 746:
```suggestion
        std::string service = _service;
        std::string     topic = "/"s + _service + "?" + request.paramsString;
```
This is what I would expect here.

wirew0rm: But this might need other changes since we do not want to include the long polling params into the topic... so I'd just merge this as is to keep giulio working on the UI part and wait for this to be resolved separately. Just to keep in mind that/how the mechanics might change and that we cannot yet rely on query parameters working over the rest backend.

wirew0rm (approved):
Merging this even though it does not address all issues, which will be resolved separately to not block other tasks.

src/majordomo/include/majordomo/RestBackend.hpp
giucam: It is still not quite clear to me what the roles of all the various parts of an uri are, so i tried to change this code as little as possible, but to me your suggested change looks like it makes sense. Maybe @ivan-cukic can chime in?

PR #275, opened 2023-04-20
More subscription+REST fixes

giucam:
These changes make subscriptions work fine for me with RestBackend. I can now subscribe to "/DeviceName/Acquisition?channelNameFilter=saw" and i correctly receive the saw wave every 2 seconds as expected.

Ralph commented on chat:
> Sending through the pub socket should be default. Sending subscriptions through the router sockets is legacy/inefficient as it duplicates traffic for each client. Sending via the ROUTER-DEALER socket pair requires a unique id for each connection/client while the subsciotion requires only a single service/worker type id.

So that probably needs some looking into, I don't know why the router socket is used instead of the pub one. However this is enough for me to go forward with the UI work.

### DIFF
--- a/src/majordomo/include/majordomo/Broker.hpp
+++ b/src/majordomo/include/majordomo/Broker.hpp
@@ -524,6 +524,7 @@ private:
             it->second--;
             if (it->second == 0) {
                 zmq_invoke(zmq_setsockopt, _subSocket, ZMQ_UNSUBSCRIBE, topic.str().data(), topic.str().size()).assertSuccess();
+                _subscribedTopics.erase(it);
             }
         }
     }
--- a/src/majordomo/include/majordomo/Broker.hpp
+++ b/src/majordomo/include/majordomo/Broker.hpp
@@ -685,25 +685,24 @@ private:
     }
 
     void dispatchMessageToMatchingSubscribers(BrokerMessage &&message) {
-        const auto topicURI               = URI<RELAXED>(std::string(message.topic()));
-        const auto it                     = _subscribedClientsByTopic.find(topicURI);
-        const auto hasRouterSubscriptions = it != _subscribedClientsByTopic.end();
+        const auto topicURI = URI<RELAXED>(std::string(message.topic()));
 
         // TODO avoid clone() for last message sent out
-        for (const auto &[topic, clientId] : _subscribedTopics) {
+        for (const auto &[topic, _] : _subscribedTopics) {
             if (_subscriptionMatcher(topicURI, topic)) {
                 // sends notification with the topic that is expected by the client for its subscription
                 auto copy = message.clone();
                 copy.setSourceId(topic.str(), MessageFrame::dynamic_bytes_tag{});
                 copy.send(_pubSocket).assertSuccess();
-            }
-        }
 
-        if (hasRouterSubscriptions) {
-            for (const auto &clientId : it->second) {
-                auto copy = message.clone();
-                copy.setSourceId(clientId, MessageFrame::dynamic_bytes_tag{});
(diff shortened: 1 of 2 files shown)


src/majordomo/include/majordomo/Broker.hpp
```diff
@@ -684,25 +685,24 @@ class Broker {
...
-        const auto hasRouterSubscriptions = it != _subscribedClientsByTopic.end();
+        const auto topicURI = URI<RELAXED>(std::string(message.topic()));
 
         // TODO avoid clone() for last message sent out
         for (const auto &[topic, clientId] : _subscribedTopics) {
             if (_subscriptionMatcher(topicURI, topic)) {
                 // sends notification with the topic that is expected by the client for its subscription
                 auto copy = message.clone();
-                copy.setSourceId(topic.str(), MessageFrame::dynamic_bytes_tag{});
+                copy.setSourceId(topic.str(), MessageFrame::dynamic_bytes_tag{}); // Is this correct??
```
wirew0rm, line 695: It is correct, the PUB socket uses the first Frame to set the message topic (while the router socket uses it to determine which client the message should be routed to).

src/majordomo/include/majordomo/Broker.hpp
```diff
@@ -684,25 +685,24 @@ class Broker {
     }
 
     void dispatchMessageToMatchingSubscribers(BrokerMessage &&message) {
-        const auto topicURI               = URI<RELAXED>(std::string(message.topic()));
-        const auto it                     = _subscribedClientsByTopic.find(topicURI);
-        const auto hasRouterSubscriptions = it != _subscribedClientsByTopic.end();
+        const auto topicURI = URI<RELAXED>(std::string(message.topic()));
 
         // TODO avoid clone() for last message sent out
         for (const auto &[topic, clientId] : _subscribedTopics) {
```
wirew0rm, line 691:
``` suggestion
        for (const auto &[topic, _] : _subscribedTopics) {
```
The value of the _subscribedTopics map is not the client id but the number of subscribed clients, to allow removing unused subscriptions.

wirew0rm (approved):
Looks good, could you  justremove the question in the code comment (or add an explanation instead).

I also looked at why the RestBackend does not use the subscription and have an initial fix.

src/majordomo/include/majordomo/Broker.hpp
giucam: Ok, i'll remove it.

src/majordomo/include/majordomo/RestBackend.hpp
```diff
@@ -848,6 +854,18 @@ struct RestBackend<Mode, VirtualFS, Roles...>::RestWorker {
...
     }
+
+private:
+    void addParameters(const httplib::Request &request, URI<>::UriFactory &uri) {
+        for (const auto &[key, value] : request.params) {
+            if (key == "LongPollingIdx") {
+                // This parameter is not passed on, it just means we
+                // want to use long polling -- already handled
+            } else {
+                uri = std::move(uri).addQueryParameter(key, value);
```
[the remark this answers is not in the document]
giucam: Actually that does not seem to be a problem at all with the router socket.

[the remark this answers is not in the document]
wirew0rm: ah yes, it actually uses the subscription topic used for subscription for the notification, so they are identical and the SubscriptionMatcher implements the logic for query parameters etc.

That means the previous behavior was because the topic handling (for the legacy subscriptions) had bugs and would notify the wrong topic which would only work for special combinations of requests and context information.

PR #280, opened 2023-04-27
Use the correct filters on the matchers for the subscriptions

giucam:
Each subscription should have a different matcher with the correct filters.

### DIFF
--- a/src/majordomo/include/majordomo/Broker.hpp
+++ b/src/majordomo/include/majordomo/Broker.hpp
@@ -265,7 +265,13 @@ private:
     Timestamp                                               _heartbeatAt = Clock::now() + settings.heartbeatInterval;
     SubscriptionMatcher                                     _subscriptionMatcher;
     std::unordered_map<URI<RELAXED>, std::set<std::string>> _subscribedClientsByTopic; // topic -> client IDs
-    std::unordered_map<URI<RELAXED>, int>                   _subscribedTopics;         // topic -> subscription count
+    struct Sub {
+        Sub(int c)
+            : count(c) {}
+        int                 count;
+        SubscriptionMatcher matcher;
+    };
+    std::unordered_map<URI<RELAXED>, Sub>                   _subscribedTopics; // topic -> subscription count and matcher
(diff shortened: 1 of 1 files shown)


src/majordomo/include/majordomo/Broker.hpp
```diff
@@ -512,17 +518,22 @@ class Broker {
 private:
     void subscribe(const URI<RELAXED> &topic) {
         auto [it, inserted] = _subscribedTopics.try_emplace(topic, 0);
-        it->second++;
-        if (it->second == 1) {
+        it->second.count++;
+        const auto &query = topic.queryParamMap();
+        for (const auto &[name, value] : query) {
+            it->second.matcher.template addFilter<opencmw::DomainFilter<std::string_view>>(name);
```
wirew0rm, line 524: This kind of defeats the purpose of the matcher being configurable, because if we just compare all query entries by adding all present queries for equality, we could just compare the URIs or use a generic match function.

The relevant matches for the filter context are defined by the worker while the broker only dispatches based on the serviceName. Therefore the Broker should rely on notifications being to the exact same topic as the subscription requested with partial/more complex matches being performed in the worker (which also already has its own subscription matcher, which would have to be extended to default include the RequestContext fields/allow the user to add custom matchers in the worker's constructor).

So while this works as a workaround, allows to test other parts of the system now and illustrates the problem nicely, I think from a design perspective this is probably a step in the wrong direction and I would prefer to merge a proper solution.

wirew0rm: ok, while trying to find an alternative I found out how the subscriptionMatcher is supposed to be used: https://github.com/fair-acc/opencmw-cpp/blob/0e54a29a8979eb1bf8927286f1863c22c40e40a3/src/majordomo/test/majordomoworker_tests.cpp#L109

This should be possible to get everything working without this PR.

There's still a few cases where I think it would be easier if this was transparent to the Broker (e.g. if the worker is not the same process as the broker and therefore cannot simply register its types in the broker.)

giucam:
Indeed, that works. Closing this.

PR #281, opened 2023-04-28
Frank/iobuffer fixes

frankosterfeld:


RalphSteinhagen (approved):
Thanks for spotting and fixing this bug! :+1:

PR #284, opened 2023-05-04
Frank/core fixes

frankosterfeld:
Fixes for IoBuffer (my own regression, sort of), and URI

Maybe URI shouldn't store string_view members, but string-independent (pos, size) pairs to indicate the substrings, and create the string_view's in the getters. That'd make assignment simpler and more robust.

### DIFF
--- a/src/core/include/URI.hpp
+++ b/src/core/include/URI.hpp
@@ -98,6 +98,9 @@ public:
         if (this == &other) {
             return *this;
         }
+
+        const char *oldBegin = other._str.data();
+
         std::swap(_str, other._str);
         std::swap(_scheme, other._scheme);
         std::swap(_authority, other._authority);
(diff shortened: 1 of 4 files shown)


wirew0rm (approved):
Nice catches, thanks for the fixes and testcases. :+1:

src/core/include/URI.hpp
```diff
@@ -98,19 +98,35 @@ class URI {
         if (this == &other) {
             return *this;
         }
+
+        const char *oldBegin   = other._str.data();
+
+        auto        adjustView = [this, oldBegin](std::string_view &view) {
+            view = std::string_view(_str.data() + std::distance(oldBegin, view.data()), view.size());
```
wirew0rm, line 105: on a second thought, this should only be a problem for short strings, when the string is interned, right? Might be better to only do the adjust_view if `*oldBegin != this.str.data()`?

frankosterfeld: Yeah, I think it only triggers with SSO. Also was thinking about such a check, will do, and maybe also extend the test a little to test more getters.

wirew0rm (approved):
:+1:  perfect, thank you.

PR #287, opened 2023-05-05
Frank/message refactor

frankosterfeld:
Replace majordomo::MdpMessage usages with mdpMessage

Added sourceId for broker usage, changed serviceName back to std::string, as it is also used for the client source ID (for broker <-> worker communication).
The send/receive functionality and ZMQ MessageFrame have been moved to a "zmq" sublibrary, which majordomo (broker/worker) depends on, as well as parts of client/ (Client.hpp), but not the RestClient.


src/zmq/include/zmq/Message.hpp
RalphSteinhagen, line 1: This file -- if ensured that the functionally is fully replaced by the ZeroMQ-agnostic `mdp::Message` -- should be deleted.

RalphSteinhagen (changes requested):
For my part, everything looks good. :+1: 

I appreciate the changes made in the naming convention from `XXX_structname` to `XXX::structname`. 
We should ensure that the namespace/using definitions are not present in the root namespace, except for examples or end-user programs.

I have two questions/suggestions:

  1. the commits that are thematically related should be merged/squashed.
  2. it seems that the `src/zmq/include/zmq/Message.hpp` file is not being used, or at least not extensively. I believe it would be okay to delete it. We don't need to keep old code at this point.
  
  Other than that, this looks good and could be merged.

src/zmq/include/zmq/Message.hpp
frankosterfeld: This file now has the code that sends/receives mdp::Message, i.e. the zmq-specific code when dealing with mdp::Message in Broker/Worker/[Zmq]Client. I also kept the MessageFrame as there are cases where single frames instead of full messages are sent/received, mainly for subscription handling.

frankosterfeld:
> For my part, everything looks good. 👍
> 
> I appreciate the changes made in the naming convention from `XXX_structname` to `XXX::structname`. We should ensure that the namespace/using definitions are not present in the root namespace, except for examples or end-user programs.
> 
> I have two questions/suggestions:
> 
> 1. the commits that are thematically related should be merged/squashed.

I'd tend to squash everything when merging. In the history, there's a separation in porting broker/worker/clients, but the generic moving/renaming is crosscutting and maybe not deserving its own commit in the history. Also the restyle bot commits (whose results differ quite a bit from my local clang-format 14 results) would have to be applied commit by commit to make it clean.

> 2. it seems that the `src/zmq/include/zmq/Message.hpp` file is not being used, or at least not extensively. I believe it would be okay to delete it. We don't need to keep old code at this point.

See other comment, this is not the old Message class, but the free functions receiving/sending mdp::Messages, and the MessageFrame RAII type, used by some low-level usages that do not send/receive full messages (subscription handling).

> 
> Other than that, this looks good and could be merged.

PR #294, opened 2023-05-16
Minor build system fixes

vimpostor:
This fixes two small problems I noticed in the cmake build system:

- Fix `WARNING` usage inside `message()`. Previously `WARN` was used, which doesn't actually exist as a `message()` mode, so instead it was silently prefixed to the string argument.
- Avoid cherry-picking patches as a different user: Some people like to gpg-sign their commits by default. If we let cmake commit with a different user, this will cause the commit to fail, because no GPG key matches the foreign address. To fix this, instead just apply the patch without committing which is also more in line with how distros would usually patch software (i.e. using the equivalent of `patch -Np1`). This also simplifies that cmake code a bit, as we no longer need to set a git user to accomodate people that haven't configured a git user yet.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,7 @@ elseif (EMSCRIPTEN)
     # clang does not work with stdlib++ ranges, we have to use libc++
     # add_compile_options(-stdlib=libc++)
     # add_link_options(-stdlib=libc++ -lc++abi)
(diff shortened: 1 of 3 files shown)


wirew0rm (approved):
Looks good, thanks. I didn't know about the no-commit flag for cherry-pick, that's indeed a lot cleaner especially if the IDE starts to reconfigure the project for multiple configurations in the background, thanks.

PR #297, opened 2023-05-30
REST - parse the requests body

ablepharus:
Read out request body.

- [x] the Content-Type is still not taken into account


src/majordomo/include/majordomo/RestBackend.hpp
```diff
@@ -611,6 +611,8 @@ struct RestBackend<Mode, VirtualFS, Roles...>::RestWorker {
 
         auto        uri = URI<>::factory();
         std::string bodyOverride;
+        std::string contentType;
+        int contentLength;
```
frankosterfeld, line 615: Initialize to 0?

wirew0rm:
will be solved by #300, thanks for pointing this out and fixing it!

wirew0rm (approved):
Thanks for the fix, looks good :+1:

Issue #243, opened 2022-12-06
[0pts] Fix starting and stopping threads in ThreadPool


ivan-cukic:
`.joinable()` is not a proper test to see whether a thread is finished or not.

Add a mechanism to check whether a thread is actually finished,
and destroy unused `jthread`s so that the following test passes:

```
TEST_CASE("ThreadPool: Thread count tests", "[ThreadPool][MinMaxThreads]") {
    struct bounds_def {
        std::uint32_t min, max;
    };
    std::array<bounds_def, 8> bounds{
        bounds_def{ 1, 1 },
        bounds_def{ 1, 2 },
        bounds_def{ 1, 4 },
        bounds_def{ 2, 2 },
        bounds_def{ 2, 4 },
        bounds_def{ 2, 8 },
        bounds_def{ 4, 4 },
        bounds_def{ 4, 8 },
    };

    for (const auto [minThreads, maxThreads] : bounds) {
        for (const auto taskCount : { 2, 4, 8, 16, 32 }) {
            fmt::print("## Test with min={} and max={} and taskCount={}\n", minThreads, maxThreads, taskCount);
            std::atomic<int> counter{ 0 };

            // Pool with min and max thread count
            opencmw::BasicThreadPool<opencmw::IO_BOUND> pool("count_test", minThreads, maxThreads);
            pool.waitUntilInitialised();

            for (int i = 0; i < taskCount; ++i) {
                pool.execute([&counter] {
                    std::this_thread::sleep_for(std::chrono::seconds(1));
                    std::atomic_fetch_add(&counter, 1);
                    counter.notify_all();
                });
            }
            REQUIRE(pool.numThreads() >= minThreads);
            REQUIRE(pool.numThreads() == std::min(std::uint32_t(taskCount), maxThreads));

            for (int i = 0; i < taskCount; ++i) {
                counter.wait(i);
                REQUIRE(pool.numThreads() >= minThreads);
                REQUIRE(pool.numThreads() <= maxThreads);
            }

            // We should have gotten back to minimum
            std::this_thread::sleep_for(std::chrono::seconds(1));
            REQUIRE(pool.numThreads() == minThreads);
            REQUIRE(counter == taskCount);
        }
    }
}
```

RalphSteinhagen:
fixed in graph-prototype by @wirew0rm 
  * https://github.com/fair-acc/graph-prototype/pull/67
  
 We should transfer this also to here (opencmw-cpp).

PR #298, opened 2023-06-01
Threadpool improvements

wirew0rm:
Backport of changes from graph prototype. See original issue:
https://github.com/fair-acc/graph-prototype/pull/67

Some noteworthy changes:
- emscripten compatibility: std::jthread -> std::thread, ranges usages
- fix race condition on thread shutdown
- fix https://github.com/fair-acc/opencmw-cpp/issues/243 (starting and stopping threads in ThreadPool)

Also enables execution of emscripten unittests via ctest and unbreak failing emscripten tests.

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -80,9 +80,9 @@ jobs:
         git clone https://github.com/emscripten-core/emsdk.git
         cd emsdk
         # Download and install the latest SDK tools.
-        ./emsdk install 3.1.30 # latest = 3.1.32
+        ./emsdk install releases-03ecb526947f6a3702a0d083083799fe410d3893-64bit
         # Make the "latest" SDK "active" for the current user. (writes .emscripten file)
-        ./emsdk activate 3.1.30
+        ./emsdk activate releases-03ecb526947f6a3702a0d083083799fe410d3893-64bit
         # Activate PATH and other environment variables in the current terminal
         source ./emsdk_env.sh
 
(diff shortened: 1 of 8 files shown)


RalphSteinhagen (approved):
Linked to https://github.com/fair-acc/graph-prototype/pull/67

Thanks for enabling the WASM test work and following up on the ThreadPool adjustments. Nice!

wirew0rm:
Added `-fprofile-update=atomic` to the coverage compiler options, see documentation: https://gcc.gnu.org/onlinedocs/gcc-9.3.0/gcc/Instrumentation-Options.html, to prevent the problem of negative coverage line hits due to concurrent coverage counter updates as described in the issue linked from the error message: 
```
 gcovr.gcov_parser.NegativeHits: Got negative hit value in gcov line 'branch  1 taken -1' caused by a bug in gcov tool, see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=68080. Use option --gcov-ignore-parse-errors with a value of negative_hits.warn, or negative_hits.warn_one.
```
https://github.com/fair-acc/opencmw-cpp/actions/runs/5142318355/jobs/9257131141#step:14:1360

Since this is not deterministic, I'm not completely sure that this will fix the issue. According to the documentation it gcc should automatically select `atomic` if `-lpthread` is used, but is not completely clear about whether this applies generally or only for `-fprofile-update=maybe-atomic`. If it reoccurs in the future the other possible problem would be unjoined threads which break coverage.

PR #302, opened 2023-06-23
cmake: bump fmt to 10.0.0

wirew0rm:
Bumps fmt to the 10.0.0 release to be on the same version as `graph-prototype`.


RalphSteinhagen (approved):
Thanks for following this up and bumping the fmt version! This looks good and should keep the compatibility and minimise ABI friction between OpenCMW and graph-prototype!

Well done! :+1:

PR #312, opened 2023-07-07
Client context emscripten support

wirew0rm:
PR just to trigger the CI, to see whether the emscripten errors are fixed and it can be incorporated into #300. I also added a small dns_example binary to be able to test the emscripten client against a native dns service which is not easily possible in the current testing framework.


wirew0rm:
this PR was never expected to be merged and is incorporated into #300 which has since been merged

PR #323, opened 2023-10-20
Fix subscription topic handling

frankosterfeld:
- Ensure that a client subscription "/service?a=b" arrives in the worker as such, and not as "/service".
 - Ensure that clients subscriptions "/service?a&b" and "/service?b&a" are treated the same, i.e. that the params are sorted alphabetically when passed to ZMQ.


wirew0rm (approved):
Looks good :+1: 

Note: applied the restyled changes

PR #324, opened 2023-10-23
Make sure we don't fail on double patching mpunits

ivan-cukic:
CMake has an issue of sometimes trying to apply the same PATH_COMMAND twice. [1] This patch adds a workaround to make sure git apply doesn't fail.

https://gitlab.kitware.com/cmake/cmake/-/issues/21146


RalphSteinhagen (approved):
:+1:  lgtm providing the CI passes.

Issue #329, opened 2023-11-13
[1pt, 0.5pt] Rest: Long-poll subscriptions without notifications disable HTTP interface


frankosterfeld:
When the UI creates a long-poll subscription (e.g. `/dashboards`) that receives only sporadic or no updates, the UI might run into a client-side timeout, triggering the REST client to start a new request.

The handling thread in RESTBackend.hpp remains being blocked waiting for a notification from the worker though (`connection->waitForUpdate();`). Thus each timeout/new request blocks another HTTP worker thread until the pool is exhausted. When that happens, the REST backend stops handling requests.

PR #330, opened 2023-11-23
Do not block worker threads forever if no event is received (#329)

frankosterfeld:
Make sure that long-poll request handlers do not block forever when no corresponding event is received. Otherwise the clients will send one request after another once their request times out client-side, until the worker threads are exhausted and the server stops responding.

It would be great if we could also detect the client connection be dropped using Keep-Alive, but cpp-httplib doesn't seem to have API for that.


RalphSteinhagen (changes requested):
This seems to look OK but should be accompanied by a unit-tests that demonstrate the feature/workaround.

PR #333, opened 2023-12-05
Rename cmake options from ENABLE_* to OPENCMW_ENABLE_*

ivan-cukic:


RalphSteinhagen (approved):
Thanks for adding the specific test enable/disable flag! 👍

PR #341, opened 2024-02-26
Https subscription and root certs

ablepharus:
Implement https subscriptions and fix a bug related to our handling regarding the https client.

Https subscriptions were just not handled, initialise a client and reuse the existing code for http subscriptions.
There was a bug in our certificate store handling. The SSLClient owns its certificate store and deletes it on destruction.
We can not share one store with all clients, so every client gets its own now.

### DIFF
--- a/src/client/include/RestClientNative.hpp
+++ b/src/client/include/RestClientNative.hpp
@@ -142,7 +142,6 @@ class RestClient : public ClientBase {
     std::mutex                             _subscriptionLock;
     std::map<URI<STRICT>, httplib::Client> _subscription1;
 #ifdef CPPHTTPLIB_OPENSSL_SUPPORT
-    X509_STORE                               *_client_cert_store = nullptr;
     std::map<URI<STRICT>, httplib::SSLClient> _subscription2;
 #endif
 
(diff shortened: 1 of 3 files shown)


wirew0rm (changes requested):
Looks good in general.
I think it would be good to persist the findings about how the certstore is handled. Please update the PR description and commit messages accordingly.

Also this adds quite some new uncovered new code paths which I think should be easy to fix, since RestClient_tests.cpp already has http(s) tests for get/set and a http only test for subscribe.

src/client/include/RestClientNative.hpp
```diff
@@ -285,7 +278,7 @@ class RestClient : public ClientBase {
         if (cmd.topic.scheme() && equal_with_case_ignore(cmd.topic.scheme().value(), "https")) {
 #ifdef CPPHTTPLIB_OPENSSL_SUPPORT
             httplib::SSLClient client(cmd.topic.hostName().value(), cmd.topic.port() ? cmd.topic.port().value() : 443);
-            client.set_ca_cert_store(_client_cert_store);
+            client.set_ca_cert_store(detail::createCertificateStore(_caCertificate));
```
wirew0rm, line 281: Maybe even a short comment here why every client should get their own cert store.

wirew0rm (approved):
Looks good, thanks for the additional documentation.

I've looked into the coverage, it seems most of it is due to coverage not being properly configured in this project, the majority of the missed lines is in the unit-test that seems to be excluded from coverage collection but not coverage calculation :/

PR #342, opened 2024-02-28
implement deletion of dns entries and REST api for it

ablepharus:


frankosterfeld:
Context: this is needed to unregister signals in the opendigitizer service, when the signal metadata of a sink changes, or when the whole flow graph is replaced.

src/services/include/services/dns_client.hpp
```diff
@@ -90,12 +112,21 @@ struct DnsClient {
...
+    std::vector<Entry> deleteSignals(const std::vector<Entry> &entries, std::chrono::seconds timeout = 20s) {
+        return _doSync<std::vector<Entry>>(&DnsClient::deleteSignalsAsync, entries, timeout);
+    }
     Entry registerSignal(const Entry &entry) {
         auto s = registerSignals({ entry });
         if (s.size() > 0)
             return s[0];
         return {};
     }
+    Entry deleteSignal(const Entry &entry) {
```
frankosterfeld, line 124: Thinking about it, I'd prefer `unregisterSignal(s)`. It's symmetric with registerSignal(s), and also sounds less threatening :)

frankosterfeld:
besides unregister vs. delete wording preference, looks good to me

src/services/include/services/dns_storage.hpp
```diff
@@ -60,6 +60,30 @@ class DataStorage {
...
+        });
+        return deletedEntries;
+    }
+
+    template<class ReturnEntryType = EntryType, class FilterT = FilterType>
+    std::vector<ReturnEntryType> deleteEntries(const FilterT &filter = {}) {
+        std::vector<ReturnEntryType> deletedEntries;
+        std::erase_if(_entries,
+                [&filter, &deletedEntries](const StorageEntryType &entry) {
+                    if (filter == entry) {
```
wirew0rm, line 78: Do we really want to check for full equality here and not just delete by name?

ablepharus: I could change the default type to QueryEntry, that way you could delete by any field, or pass the exact entry.

wirew0rm (approved):

PR #354, opened 2024-10-18
Disruptor: Correctly rethrow exception

frankosterfeld:
Don't rethrow the exception_ptr, but the contained exception.


wirew0rm (approved):
looks good, can be merged once CI passes.

PR #355, opened 2024-10-31
Upgrade to cpp-httplib v0.18.1

dantti:
Enabling TESTING on opendigitizer fails
to build as due httplib used here not
having the required status enums.


RalphSteinhagen (approved):
Thanks for bumping! CI passed and LGTM. :+1:

PR #359, opened 2024-12-03
Implement RBAC integration

vimpostor:
Please refer to #349 and to the commit message for more info.


RalphSteinhagen (approved):
@vimpostor thanks for the core implementation. Looks good as is.

Next step for us would be to integrate the different role and hash checks in the different worker implementation (another to be created issue).

RalphSteinhagen:
@vimpostor thanks for adding the cmake entry for lib sodium! :+1:

PR #362, opened 2025-02-12
Rest Subscription Improvements

wirew0rm:
# Proxy fetch from worker threads to main thread

Fetch calls on pthreads will only be handled once the thread terminates and the control of the underlying web-worker is given back to the emscripten runtime, which most of the time never happens as they belong to a static thread-pool.

The fact that most of the subscriptions worked before was because the start handler or settings_updated handler was not called by the block worker, but by the main thread setting up the graph.

For some reason it was necessary to increase the stack size for this proxying to work.

# Rest Backend
- bump httplib version
- set default settings
- redirect to newest instead of errors for longpolling request to outdated data

# minor codecleanups
see commit messages of individual commits

### DIFF
--- a/cmake/DependenciesNative.cmake
+++ b/cmake/DependenciesNative.cmake
@@ -1,4 +1,4 @@
-# Build a static version of openssl to link into opencmw
+# Build a static version of openssl to link into
 set(OPENSSL_INSTALL_DIR "${CMAKE_BINARY_DIR}/openssl-install")
 add_library(OpenSSL::Crypto STATIC IMPORTED GLOBAL)
 add_library(OpenSSL::SSL STATIC IMPORTED GLOBAL)
@@ -38,6 +38,9 @@ FetchContent_Declare(
         GIT_TAG v4.3.4 # latest v4.3.4
 )
 set(ZMQ_BUILD_TESTS OFF CACHE BOOL "Build the tests for ZeroMQ")
+# suppress warnings for missing zeromq dependencies by disabling some features
+set(WITH_TLS OFF CACHE BOOL "TLS support for ZeroMQ WebSockets")
+set(BUILD_SHARED OFF CACHE BOOL "Build cmake shared library")
 option(WITH_PERF_TOOL "Build with perf-tools" OFF)
 
 FetchContent_Declare(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.19)
 
-project(opencmw-cpp CXX)
+project(opencmw-cpp CXX C) # C is needed to compile the internal version of libsodium
 set(CMAKE_CXX_STANDARD 23)
 
 # include(cmake/StandardProjectSettings.cmake)
--- a/cmake/CMakeRC.cmake
+++ b/cmake/CMakeRC.cmake
@@ -34,7 +34,7 @@ endif()
 
 set(_version 2.0.0)
 
-cmake_minimum_required(VERSION 3.3)
+cmake_minimum_required(VERSION 3.19)
 include(CMakeParseArguments)
 
 if(COMMAND cmrc_add_resource_library)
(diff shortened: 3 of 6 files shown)


cmake/DependenciesNative.cmake
```diff
@@ -26,16 +26,10 @@ FetchContent_Declare(
         GIT_TAG openssl-3.0.5 # latest openssl-3.0.5
 )
 
-# mustache is forked into 3rd_party/
```
RalphSteinhagen, line 29: OpenSSL is presently an unimported FetContent target. Rather than deleting it, please add it as a static dependency. 

This way we can better control for which compiler optimisations, architecture (very important that at least AVX2 <-> simd), and version is being used. This not only for performance reasons but also security auditability where the whole OpenCMW including all its immediate and transitory dependencies need to be verifiable.

See for example: https://stackoverflow.com/a/72187679/2411625

Or specifically (O1-generated):
```cmake
cmake_minimum_required(VERSION 3.20)
project(MyProject LANGUAGES C CXX)

include(FetchContent)

set(OPENSSL_INSTALL_DIR "${CMAKE_BINARY_DIR}/openssl-install")

# Fetch OpenSSL
FetchContent_Declare(
    openssl
    GIT_REPOSITORY https://github.com/openssl/openssl.git
    GIT_TAG openssl-3.0.0  # Change this to the desired OpenSSL version
)

FetchContent_GetProperties(openssl)
if(NOT openssl_POPULATED)
    FetchContent_Populate(openssl)

    # Ensure we use specific optimisations for OpenSSL compilation
    set(OPENSSL_C_FLAGS "-O3 -march=native" CACHE STRING "OpenSSL custom CFLAGS" FORCE)
    set(OPENSSL_CXX_FLAGS "-O3 -march=native" CACHE STRING "OpenSSL custom CXXFLAGS" FORCE)
    
    # Manually configure and build OpenSSL
    execute_process(
        COMMAND ./Configure no-shared no-tests --prefix=${OPENSSL_INSTALL_DIR} --openssldir=${OPENSSL_INSTALL_DIR} linux-x86_64
        WORKING_DIRECTORY ${openssl_SOURCE_DIR}
        ENV "CFLAGS=${OPENSSL_C_FLAGS}" "CXXFLAGS=${OPENSSL_CXX_FLAGS}"
    )
    execute_process(
        COMMAND make -j
        WORKING_DIRECTORY ${openssl_SOURCE_DIR}
    )
    execute_process(
        COMMAND make install_sw
        WORKING_DIRECTORY ${openssl_SOURCE_DIR}
    )
endif()

# Create "imported" static library targets for OpenSSL
add_library(OpenSSL::Crypto STATIC IMPORTED)
add_library(OpenSSL::SSL STATIC IMPORTED)

# Ensure OpenSSL is built before our targets
add_dependencies(OpenSSL::Crypto openssl)
add_dependencies(OpenSSL::SSL openssl)

# Set the library and include paths
set_target_properties(OpenSSL::Crypto PROPERTIES
  IMPORTED_LOCATION "${OPENSSL_INSTALL_DIR}/lib/libcrypto.a"
  INTERFACE_INCLUDE_DIRECTORIES "${OPENSSL_INSTALL_DIR}/include"
)
set_target_properties(OpenSSL::SSL PROPERTIES
  IMPORTED_LOCATION "${OPENSSL_INSTALL_DIR}/lib/libssl.a"
  INTERFACE_INCLUDE_DIRECTORIES "${OPENSSL_INSTALL_DIR}/include"
)

# Example: Statically linking OpenSSL into your executable
add_executable(MyApp main.cpp)
target_link_libraries(MyApp
  PRIVATE
    OpenSSL::SSL
    OpenSSL::Crypto
)
```

RalphSteinhagen (approved):
Thanks for the changes. If you could address the OpenSSL item mentioned above. This could be then merged as is.

cmake/DependenciesNative.cmake
wirew0rm: Our case is a little bit different, because openssl is a transitive dependency (via httplib), but it seems to be sufficient that the OpenSSL::* targets exist prior to fetching the libhttp dependency. also Fetch content populate is deprecated.
I got it almost working, it uses the statically compiled version now, but it will recompile it everytime, breaking incremental compilation.

wirew0rm: fixed, along with some other cmake issues. Will merge once the CI has passed.

PR #366, opened 2025-03-26
Reimplement REST interface using HTTP/2

frankosterfeld:
Instead of using the httplib-based HTTP server using blocking I/O in a thread per client, implement a nghttp2-based HTTP/2 server, using single-threaded non-blocking I/O directly integrated into the Broker zmq_poll() loop.


frankosterfeld:
Spamming CI, will reopen later

PR #371, opened 2025-04-24
Reimplement REST interface using HTTP/2, HTTP3

frankosterfeld:
Implements a new REST interface using HTTP/2, for lower latency and multiplexing of multiple requests on a single connection.
Instead of multithreading + blocking I/O, where each (long-poll) client occupies a thread, using event-driven non-blocking I/O, and poll() the HTTP-related sockets (server socket and client connections) together with the ZMQ sockets. For that, integrate the server more closely with the Broker.

The HTTP/3 implementation has been tested manually using Chrome and curl (see README) but is not covered by unit tests yet, as the native HTTP/3 client implementation would be considerable effort (and probably some more kloc's of code...). As agreed upon with @RalphSteinhagen , the native client might be done later possibly with other libraries like libcurl/CPR that require less code than ngtcp2, given that the client side is not the performance-critical part.


cmake/DependenciesNative.cmake
```diff
@@ -1,32 +1,113 @@
-# Build a static version of openssl to link into
+include(ExternalProject)
+
+set(OPENSSL_C_FLAGS "-O3 -march=x86-64-v3" CACHE STRING "OpenSSL custom CFLAGS" FORCE)
```
RalphSteinhagen, line 3: `-march=x86-64-v3` should not be hard-coded...

cmake/DependenciesNative.cmake
```diff
@@ -1,32 +1,113 @@
...
+add_library(nghttp2-static STATIC IMPORTED GLOBAL)
+set_target_properties(nghttp2-static PROPERTIES
+        IMPORTED_LOCATION "${CMAKE_BINARY_DIR}/nghttp2-install/lib/libnghttp2.a"
+        INTERFACE_INCLUDE_DIRECTORIES "${CMAKE_BINARY_DIR}/nghttp2-install/include"
 )
-add_custom_target(openssl-build ALL
-        DEPENDS ${OPENSSL_INSTALL_DIR}/lib64/libcrypto.a ${OPENSSL_INSTALL_DIR}/lib64/libssl.a
+add_dependencies(nghttp2-static Nghttp2Project)
+
+ExternalProject_Add(Nghttp3Project
```
RalphSteinhagen, line 58: same here....

concepts/majordomo/assets/mustache/ServicesList.mustache
```diff
@@ -96,7 +96,7 @@ input.topicInput { border: 0; width: 100px; border-bottom: 1px solid silver; }
             };
 
             listenButton.onclick = () => {
-                        let post = { method: 'GET', headers: { 'X-OPENCMW-METHOD' : 'POLL' } };
+                        let post = { method: 'GET', headers: { 'x-opencmw-method' : 'POLL' } };
```
RalphSteinhagen, line 99: Why do we still need this 'secret sauce' header field?

frankosterfeld: Ah, bodyOverride is also used by our RestClients, of course. I'll try to remove those.

wirew0rm: iirc these headers were there in the very beginning, before we introduced the `longPollingIdx` query parameter to differentiate between the different mdp methods that would get mapped to http get, but would need different handling on the server side.
Apart from maybe the mustache part (which I'm not that familliar with), I don't think it's really required for the current logic.

src/client/include/RestClientEmscripten.hpp
```diff
@@ -175,21 +189,28 @@ struct SubscriptionPayload : FetchPayload {
...
+
+                char       *end                  = nullptr;
+                longPollingIdx                   = strtoull(longPollingIdxString.data(), &end, 10);
+                if (end != longPollingIdxString.data() + longPollingIdxString.size()) {
+                    std::println(std::cerr, "RestClientEmscripten::payloadError: url: {}, bytes: {}\n", fetch->url, fetch->numBytes);
+                    return;
+                }
+
                 if (payload->_update != 0 && longPollingIdx != payload->_update) {
                     std::print("received unexpected update: {}, expected {}\n", longPollingIdx, payload->_update);
```
drslebedev, line 207: Clients report "unexpected update" errors, with the received index always off by one:
```text
received unexpected update: 7, expected 6    index.html:80:37  
received unexpected update: 66, expected 65  index.html:80:37  
received unexpected update: 67, expected 66  index.html:80:37  
received unexpected update: 7, expected 6    index.html:80:37  
received unexpected update: 14, expected 13  index.html:80:37  
```

It looks like this error appears to be caused by incorrect checks. There’s no indication of actual data loss, and visually everything looks fine in ImPlot. Each time the error occurs, a WARNING_SAMPLES_DROPPED tag is published, but the curve remains smooth, with no apparent data gaps. Most likely, it’s a simple off-by-one indexing error.
Screenshot from 2025-06-20 11-04-48

drslebedev: It looks like the reason is that SubscriptionPayload::request() compares the new longPollingIdx against the last value already processed (_update). With sequential data the very next payload will always be _update + 1.

```cpp
                if (payload->_update != 0 && longPollingIdx != payload->_update + 1) {
                    std::print("received unexpected update: {}, expected {}\n", longPollingIdx, payload->_update + 1);
                }
                payload->onsuccess(fetch->status, std::string_view(fetch->data, detail::checkedStringViewSize(fetch->numBytes)), static_cast<long>(longPollingIdx) - static_cast<long>(payload->_update + 1));
```

cmake/DependenciesNative.cmake
```diff
@@ -1,32 +1,113 @@
...
+        -DENABLE_HTTP3:BOOL=OFF
+        -DENABLE_DEBUG:BOOL=${ENABLE_NGHTTP_DEBUG}
+        -DBUILD_STATIC_LIBS:BOOL=ON
+        -BUILD_SHARED_LIBS:BOOL=OFF
+        -DENABLE_DOC:BOOL=OFF
+)
+
+add_library(nghttp2-static STATIC IMPORTED GLOBAL)
+set_target_properties(nghttp2-static PROPERTIES
+        IMPORTED_LOCATION "${CMAKE_BINARY_DIR}/nghttp2-install/lib/libnghttp2.a"
```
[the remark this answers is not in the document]
wirew0rm: i've added a simple commit to fix the lib dir problems to be able to test and deploy this, so this is resolved.
Incremental compilation is fixed by adding an empty update command to the ExternalProject call.

wirew0rm (approved):
Thanks for this big improvement. We did some real-world testing and it improved a lot of long-standing issues, thanks a lot :+1:
There are still some small issues and we have to rethink how to handle dependencies, but this can be tackled in individual small PRs.

PR #375, opened 2025-05-22
Replace fmt with std::format, std::print

drslebedev:


concepts/core/BasicThreadPool_example.cpp
```diff
@@ -19,21 +15,21 @@ int main() {
     assert(poolIO.isInitialised());                                               // check if the pool is initialised
 
     // enqueue and add task to list -- w/o return type
-    poolWork.execute([] { fmt::print("Hello World from thread '{}'!\n", getThreadName()); });
-    // poolWork.execute([]<typename... T>(T&&...args) { fmt::print(fmt::format_string<T...>("Hello World from thread '{}'!\n"), std::forward<T>(args)...); }, getThreadName());
-    // poolWork.execute([](const auto &...args) { fmt::print(fmt::runtime("Hello World from thread '{}'!\n"), args...); }, getThreadName());
+    poolWork.execute([] { std::print("Hello World from thread '{}'!\n", getThreadName()); });
+    // poolWork.execute([]<typename... T>(T&&...args) { std::print(std::format_string<T...>("Hello World from thread '{}'!\n"), std::forward<T>(args)...); }, getThreadName());
+    // poolWork.execute([](const auto &...args) { std::print(std::runtime("Hello World from thread '{}'!\n"), args...); }, getThreadName());
```
wirew0rm, line 20: `std::runtime` should probably be `std::runtime_format` and is only available in c++26. not important since it's only in comment, but probably this also illustrates why we shouldn't have commented out code ;)

src/majordomo/include/majordomo/Broker.hpp
```diff
@@ -65,14 +66,14 @@ struct fmt::formatter<opencmw::majordomo::detail::DnsServiceItem> {
...
-        return fmt::format_to(ctx.out(), "[{}: {}]", v.serviceName, fmt::join(v.uris | std::views::transform(opencmw::majordomo::detail::uriAsString), ","));
+        return std::format_to(ctx.out(), "[{}: {}]", v.serviceName, opencmw::join(v.uris | std::views::transform(opencmw::majordomo::detail::uriAsString), ","));
 #else
         std::vector<std::string> uris{};
         uris.reserve(v.uris.size());
         for (auto &uri : v.uris) {
             uris.emplace_back(opencmw::majordomo::detail::uriAsString(uri));
         }
-        return fmt::format_to(ctx.out(), "[{}: {}]", v.serviceName, fmt::join(uris, ","));
+        return std::format_to(ctx.out(), "[{}: {}]", v.serviceName, std::join(uris, ","));
```
wirew0rm, line 76:
```suggestion
        return std::format_to(ctx.out(), "[{}: {}]", v.serviceName, opencmw::join(uris, ","));
```
possible oversight, but this would mean that this branch of the file is not compiled at all? probably because broker is only compiled for GCC?

wirew0rm (approved):
Looks good, thanks for pulling this through. Found some minor things, the only one that could be a real issue is the `std::join`. Please have a look and change/merge as appropriate.

conanfile.py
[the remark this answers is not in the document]
drslebedev: I think we will remove it but in the next PRs.

PR #381, opened 2025-06-30
HTTP Server: make upgrade to http3 configurable

wirew0rm:
This by default disables the upgrade from http2 to http3, since there were some problems discovered that need to be fixed first.

By only disabling the upgrade, the functionality is still there and can be tested, but does not break for unsuspecting users.


RalphSteinhagen (approved):
Looks good! :+1: 

Please add an issue to this PR (comment) detailing what needs to be fixed for the HTTP/3 upgrade support before merging this. Thanks in advance.

PR #407, opened 2026-07-08
Implement RBAC message discarding filters

### LINKED ISSUE #403: [2SP;3SP] Integrate RBAC into GNU Radio 4 message handling
- Add message dicarding filters based on a user-defined predicate.
- Implement tests to discard/approve messages based on user roles
- the RBAC checks should be done only on the OpenCMW in/out side.. Once the messages are received there should be no further RBAC checks in GR4. I.e. shift this to the upper point.

magnusg-kdab:
In order to be flexible with the message discarding, this uses a generic predicate to be able to configure the discarding rules.

Closes #403

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -96,7 +96,7 @@ jobs:
         shell: bash
         env:
           KEYCLOAK_URL: http://keycloak:8080
-          KEYCLOAK_REDIRECT_URI: http://localhost:8091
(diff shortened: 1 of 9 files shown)


RalphSteinhagen (approved):
@magnusg-kdab thanks for the update/missing-feature fix. Looks goot.

_N.B. sorry for having kept this dangling for so long 🙈 ._

Can be merged once the CI passes.

